PR #6, opened 2020-02-25
Address the change of check runner name binary

Barthelemy:
by adding a number of tasks and workflows


graduta:
You mentioned `readout-qc-1` should never be used anymore. Could we remove it from workflow?

Barthelemy:
I did not express myself correctly. If someone uses v0.5.0 then they need to use it.

Barthelemy:
@knopers8

knopers8:
It looks like it should work, assuming that the checker name corresponds to the check names which will be running in that configuration.

graduta:
This is related to the blocker: OCTRL-189. Could we merge it so that we can test it?

Barthelemy:
I cannot merge it.

awegrzyn:
This PR broke FLP Suite CI as it uses `readout-qc-v0.5.0` during the tests phase...

awegrzyn:
I tried `readout-qc-v0.7.0` but only what I get is `DeadlineExceeded`

teo:
Ouch, looks like the rename did it. @Barthelemy @knopers8 can we make it easy and avoid this rename?

knopers8:
If you mean the file rename, then that is okay with me.

Barthelemy:
we can

Barthelemy:
https://github.com/AliceO2Group/ControlWorkflows/pull/7

PR #25, opened 2021-01-11
Emcalvs

teo:


tasks/stfb-internal-dpl-clock.yaml
```diff
@@ -1,6 +1,6 @@
 name: stfb-internal-dpl-clock
 defaults:
-  stfb_dataspec: "B:TPC/RAWDATA"
+  stfb_dataspec: "B:{{detector}}/RAWDATA"
```
teo, line 3: Please note that since these values are in the `defaults` section of a *task template*, any `stfb_dataspec` value defined as `vars` anywhere else will override it. So e.g. if the workflow template says `stfb_dataspec: B:EMCALVS/RAWDATA`, it will override this.

tasks/stfbuilder-nooutput.yaml
```diff
@@ -6,7 +6,7 @@ defaults:
   user: flp
   fmq_severity: info
   fmq_verbosity: high
-  detector: TPC
+  detector: "{{detector}}"
```
teo, line 9: Same as `stfb_dataspec`, if the workflow template (or your own user-provided variables at environment creation) has a variable definition for `detector`, it will override this without having to edit each task template.

workflows/readout-dataflow.yaml
```diff
@@ -8,7 +8,7 @@ defaults:
   monitoring_qc_url: "no-op://"
   monitoring_dd_url: "no-op://"
   monitoring_readout_url: "no-op://"
-  detector: TPC
+  detector: "{{detector}}"
```
teo, line 11: Since you've used a `{{}}` expression within the top-level `defaults` map, you have basically made `detector` a mandatory parameter, because there is now no default value to fall back to.

Basically none of these changes are necessary at all, as long as you pass `detector` and `stfb_dataspec` as runtime parameters. Alternatively, you can specify here in the top-level defaults something like:
```
detector: EMCALVS
stfb_dataspec: "B:EMCALVS/RAWDATA"
```
and then you won't have to pass any extra params.

In either case, the relevant values will be propagated to the rest of the environment, overriding any task setting.

mfasDa:
Hi @teo ,

Thanks for your review!

In order to avoid misunderstanding: Currently the detector string is hard coded to "TPC" in several config files. I want to change the detector string to "EMC" for our detector, because we are not TPC. This means also the data spec for the STFbuilder should be "EMC/RAWDATA" and not "TPC/RAWDATA". I understood that you have a template system which would replace {{detector}} with the value specified for detector in AliECS, but maybe I misunderstood something. With the current tasks and workflows the STFBuilder definitely publishes TPC/RAWDATA, no matter what we specify in AliECS for detector, and if I specify "EMC" then AliECS crashes during CONFIGURE - only if I set back the detector to "TPC" I manage to configure. If possible I would prefer to stay with the standard workflow and not copy clone the workflow and task just to change the detecotr string.

Cheers

Markus

workflows/readout-dataflow.yaml
mfasDa: I understood {{detector}} would be a variable set by AliECS, therefore it looked to me appropriate. Maybe I misunderstood something.

What I want to do is to change from hard-coded "TPC" to a variable field where we can set our detector string "EMC" via AliECS. Not every detector in ALICE is TPC.

teo: That's ok, let me clarify:
**all** keys written under a `defaults` or `vars` section can be overridden by values provided by the user, regardless of their content, either via the AliECS GUI or via `coconut`.

In other words, if you want the `detector` value to be overridden by your setting, you don't need `detector: {{detector}}`. The default specified there could very well be `detector: TEST` or whatever other dummy value, it will be overridden regardless of this default value, as long as you pass another value with the same key "`detector`". The default string `TPC` there was not intended as a slight to EMC or any other detector, it was simply a value we had used during testing, aligned with the default StfBuilder configuration IIRC.

In production you will not need to do this, as different subtrees of the main production workflow will set their own different values of `detector`.

I believe the TPC string we saw in those logs *must* come from somewhere else, not from AliECS.

teo:
Hi, you are right that there is a template system, but the StfBuilder configuration files aren't generated from this template system yet (will likely be in Suite v0.15 due in February), so a change in AliECS configuration must also currently manually be applied to StfBuilder config file(s), if/when relevant. It is up to you whether you prefer to edit them as files or as payloads in Consul (both are installed by `o2-flp-setup` and are equivalent for the time being, as we phase out the files).

workflows/readout-dataflow.yaml
mfasDa: I see, so you me we can configure in AliECS:

detector: EMC
stfb_dataspec: "EMC/RAWDATA"

then it will overwrite the settings in the task as those are just defaults? OK, I can try on our test setup.

teo: > I see, so you me we can configure in AliECS:
> 
> detector: EMC
> stfb_dataspec: "EMC/RAWDATA"
> 
> then it will overwrite the settings in the task as those are just defaults? OK, I can try on our test setup.

Exactly!
This is the order of priority of the AliECS hierarchical key-value store, from "least defaulty" i.e. "strongest" to "most defaulty" at the bottom, so basically anything above overrides anything below (for the same key):
* user-provided values
* workflow template vars, in hierarchical order
* workflow template defaults, in hierarchical order
* task vars
* task defaults
* global defaults
* built-in values

mfasDa: OK, I tried this on our test FLP stetting the two variables to detector: EMC and stfb_dataspec: "EMC/RAWDATA", and the workflow crashes because channel matching to EMC/RAWDATA fails since only TPC/RAWDATA is available. So currently the solution does not work for us.

A side comment: While for the detector experts it is of course not a problem to set two variables I wonder how the average ECS shifter will get along with this if it has to be defined for every detector separately, better limit the number of variables, i.e. if the detector is EMC (or whatever) the dataspec EMC/RAWDATA (or WHATEVER/RAWDATA) should be a consequence.

PR #27, opened 2021-01-15
Parametrize MODULEPATH against variable modulepath

teo:
OCTRL-364 #close

### DIFF
--- a/tasks/fairmq-ex-1-n-1-processor.yaml
+++ b/tasks/fairmq-ex-1-n-1-processor.yaml
@@ -14,5 +14,5 @@ command:
   user: flp
   arguments: []
   value: >-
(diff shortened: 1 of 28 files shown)


awegrzyn:
Hi @teo, just FYI, `MODULEPATH` is set globally with modules v4, so you could even drop it.

teo:
@awegrzyn I'm aware, but I don't want to. Now that we don't need any `aliswmod`/`alienv` magic, parametrizing `MODULEPATH` allows the user to point to an `aliBuild` modules directory and run dataflow components directly from built sources, skipping the RPMs step (assuming a single node setup or a multi-node setup with some other additional method of mirroring that build dir).

PR #30, opened 2021-02-04
Add options to DPL devices

miltalex:


workflows/readout-dataflow.yaml
```diff
@@ -10,6 +10,7 @@ defaults:
   monitoring_qc_url: "no-op://"
   monitoring_dd_url: "no-op://"
   monitoring_readout_url: "no-op://"
+  infologger_mode: "no-op://"
```
teo, line 13: I believe this to be incorrect. Please see https://github.com/AliceO2Group/InfoLogger/blob/master/doc/README.md and search for INFOLOGGER_MODE, there's a list of allowed values.

teo (changes requested):

workflows/readout-dataflow.yaml
miltalex: Yes, my bad this branch is obsolete I will close the MR and open a new one.

PR #35, opened 2021-02-17
DD parameters

ironMann:
@vascobarroso @teo I did a bit of systematization. Nothing major though.

### DIFF
--- a/tasks/stfsender.yaml
+++ b/tasks/stfsender.yaml
@@ -10,8 +10,6 @@ defaults:
   dd_discovery_endpoint: "no-op://" # location of consul docker instance
   user: flp
   fmq_severity: info
-  fmq_verbosity: high
-  detector: TPC
 command:
   env:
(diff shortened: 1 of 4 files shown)


tasks/stfsender.yaml
```diff
@@ -21,17 +19,17 @@ command:
   shell: true
   user: "{{ user }}"
   value: >-
-    source /etc/profile.d/modules.sh && MODULEPATH={{ modulepath }} module load DataDistribution Control-OCCPlugin &&
+    curl {{ dd_discovery_endpoint }} -fs -o /dev/null && source /etc/profile.d/modules.sh && MODULEPATH={{ modulepath }} module load DataDistribution Control-OCCPlugin &&
```
ironMann, line 22: We want checks like this curl cmd trying to reach DD consul. (this will fail for standalone runs)

teo (approved):

PR #43, opened 2021-03-09
Fix the data flow in QC workflow

knopers8:
This fixes two issues:
- incorrect workflow connection (Check Runner sending Qualities to STFS). Now STFS gets data from the dpl output proxy
- data was stuck in the DPL input proxy, because the headers would not match the specified data spec. This requires also https://gitlab.cern.ch/AliceO2Group/system-configuration/-/merge_requests/1307


knopers8:
Old files are deleted in a separate commit, so we can revert them if needed. I would advise against squashing this PR.

teo:
> Old files are deleted in a separate commit, so we can revert them if needed. I would advise against squashing this PR.

That's ok, I squash almost never.

teo (approved):

PR #49, opened 2021-03-15
Use fmq_rate_logging variable for rateLogging

miltalex:


teo:
This is good, I just wanted to check with @knopers8 @vascobarroso : should `fmq_rate_logging` in `readout-dataflow` default to `"10"` as it does now, or do we default to `"0"` thus making it necessary to explicitly enable it to get any channel rate logging?

teo (approved):

vascobarroso:
I think this kind of logging will still be useful, at least until we start getting these values in the monitoring system. So I'd leave it at 10 for now.

knopers8:
I agree with Vasco

PR #50, opened 2021-03-16
PHOS compressor workflow

knopers8:


knopers8:
I've included the fixes after the first session with PHOS. During the second session we confirmed that the workflow is fine. The problem reported in O2-2178 is most likely unrelated to the workflow template itself.

PR #52, opened 2021-03-25
Regenerate workflows

knopers8:
To be tested.

This includes all the workflows we tested so far regenerated with O2v21.11 and with workflow inclusion feature in mind. The TOF compressor has been regenerated according to the script provided by Francesco (two FLPs use a bit different configuration, thus two workflows).


knopers8:
All are working from the execution point-of-view, except for TOF compressors which meet the problem described here https://alice.its.cern.ch/jira/browse/OCTRL-448

vascobarroso (approved):

PR #56, opened 2021-04-08
[OCTRL-447] Documentation of automatic workflow template generation

knopers8:
@teo @miltalex @vascobarroso @Barthelemy It would be great if one of you tried to follow it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,3 +117,163 @@ These checks are ran automatically against:
 *  `walnut` version: [Control/master](https://github.com/AliceO2Group/Control/tree/master/)
 
 Source code can be found [here.](.github/workflows/template.yml)
+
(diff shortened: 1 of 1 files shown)


Barthelemy:
The script did not work on Mac. I think that it is fine but you could add a sentence about it.

knopers8:
Thanks, I will apply your comments. The script probably doesn't work on mac because of the differences in `sed`. Indeed it requires a comment at least.

knopers8:
OK, added a bit of explanation on `readout-dataflow` and a comment about `sed` on Mac.

knopers8:
Can we merge it if there are no other comments?

teo (approved):

PR #68, opened 2021-05-07
Workflow paths update for new RPMs

teo:


tasks/qc-daq-Dispatcher.yaml
```diff
@@ -1,6 +1,10 @@
 name: qc-daq-Dispatcher
 defaults:
   log_task_output: none
+  _module_cmdline: >-
+    source /etc/profile.d/modules.sh && MODULEPATH={{ modulepath }} module load O2 QualityControl Control-OCCPlugin &&
+    {{ dpl_command }} | o2-qc
+  _plain_cmdline: "source /etc/profile.d/o2.sh && {{ dpl_command }} | o2-qc"
```
knopers8, line 7: Stupid question, does `o2.sh` load also QualityControl?

knopers8:
If this is proven to work, should I support both `_module_cmdline` and `_plain_cmdline` in the future auto-generated workflows or just the new way?

tasks/qc-daq-Dispatcher.yaml
teo: From what I understand, it prepares the env vars for all of our software. @awegrzyn can correct if I'm wrong.

teo:
> If this is proven to work, should I support both `_module_cmdline` and `_plain_cmdline` in the future auto-generated workflows or just the new way?

Please support both, as this is the only way to be able to run both from new RPMs, old RPMs, and aliBuild output directory. Even when we drop support for old module-based RPMs, we'll still need this to support running from aliBuild output.

tasks/qc-daq-Dispatcher.yaml
awegrzyn: Yes, that's correct, it's a single file for all packages.

PR #79, opened 2021-06-15
DD monitoring for flp suite v0.19.0

ironMann:
@vascobarroso I think something like this is needed to send dd metric to the EPN-Grafana. But the Url should also contain the AliECS endpoint.


vascobarroso:
@ironMann I will let @teo and confirm, but I think this should be set at deployment time (i.e. Ansible) and not hardcoded in the workflow definition (otherwise all deployments - detector labs, CI, etc - will send metrics to your instance). 

@awegrzyn, how should multiple endpoints be specified ?

teo:
> @ironMann I will let @teo and confirm, but I think this should be set at deployment time (i.e. Ansible) and not hardcoded in the workflow definition (otherwise all deployments - detector labs, CI, etc - will send metrics to your instance).

Correct. The monitoring URI should be set to `no-op://` or similar in the workflow template, and we should instead add it to Ansible/Consul. @miltalex could you please ensure that `monitoring_dd_url` is defined in the relevant spot in Ansible and pushed to Consul defaults?

awegrzyn:
@vascobarroso comma separated.

vascobarroso:
Ok perfect. @miltalex, please notice that we want to define both the FLP and EPN monitoring endpoints.

ironMann:
Ok, I dropped the monitoring_dd_url definition.

miltalex:
@vascobarroso Currently we have these in Ansible and we store them in Consul
```
monitoring_qc_url: "influxdb-unix://{{ monitoring_unix_socket }}"
monitoring_dd_url: "influxdb-unix://{{ monitoring_unix_socket }}"
monitoring_dpl_url: "influxdb-unix://{{ monitoring_unix_socket }}"
```
I can override them using the inventories for P2 deployments but which values should be used?

vascobarroso:
> @vascobarroso Currently we have these in Ansible and we store them in Consul
> 
> ```
> monitoring_qc_url: "influxdb-unix://{{ monitoring_unix_socket }}"
> monitoring_dd_url: "influxdb-unix://{{ monitoring_unix_socket }}"
> monitoring_dpl_url: "influxdb-unix://{{ monitoring_unix_socket }}"
> ```
> 
> I can override them using the inventories for P2 deployments but which values should be used?

I think we only want to change `monitoring_dd_url` to, in addition to sending to `influxdb-unix://{{ monitoring_unix_socket }}` (aka telegraf), also send to the EPN monitoring endpoint as provided by @ironMann in the original commit. According to @awegrzyn, multiple endpoints should be comma separated. 

Maybe we should give it a try on the MVS and, once validated, add it in the Ansible inventories.

ironMann:
Let's follow up the connection details in O2-1815

PR #82, opened 2021-06-15
[OCTRL-498] Parallel phos workflows

Barthelemy:


Barthelemy:
Why is there an error for hmpid task that I did not touch ?

knopers8:
CI doesn't work, this is not a problem with your PR. Reviewing it now

README.md
```diff
@@ -361,7 +378,34 @@ Following this example, the full setup can be run by adding the following parame
...
+### Future improvements
 
 With the future releases we plan to allow for Just-In-Time workflow translation.
 It means that one will not have to export the templates manually anymore and they will only need to provide a file with the standalone DPL command.
+
+### Exporting worfklows for Dummies
+
+If you have read everything above, you can now follow these simplified instructions. 
+
+3. Access a FLP with the proper FLP Suite
```
knopers8, line 390:
```suggestion
3. Access an FLP with the proper FLP Suite
```

scripts/phos-compressor-raw-qcmnt3.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+
+# set -x;
+set -e;
+set -u;
+
+WF_NAME=phos-compressor-raw-qcmn-localt3
```
knopers8, line 7: I would maybe use this name instead (or even `phos-compressort3-raw-qcmn-local` because the compressor is parallelized).. This way we also leave room for some automatization in GUI, if we always keep the `*-remote` and  `*-local` convention.

This will of course imply generating the templates again.
```suggestion
WF_NAME=phos-compressor-raw-qcmnt3-local
```

scripts/phos-compressor-raw-qcmnt3.sh
```diff
@@ -0,0 +1,43 @@
...
+
+sed -i /defaults:/\ a\\\ \\\ "phos_keep_hglg: off" workflows/${WF_NAME}.yaml
+sed -i /defaults:/\ a\\\ \\\ "phos_pedestal: off" workflows/${WF_NAME}.yaml
+
+sed -i '/--pedestal/{n;s/.*/    - "{{ phos_pedestal }}"/}' tasks/${WF_NAME}-*
+sed -i '/--keepHGLG/{n;s/.*/    - "{{ phos_keep_hglg }}"/}' tasks/${WF_NAME}-*
+
+
+# QC_FINAL_CONFIG_PATH='consul-json://{{ consul_endpoint }}/o2/components/qc/ANY/any/its-qcmn-fhr-fee'
+WF_NAME=phos-compressor-raw-qcmn-remote
```
knopers8, line 32: this will overwrite the non-parallel workflow, which is fine in principle, but perhaps won't be very clear. could you add `t3` here as well (in the same convention as for "local")?

tasks/phos-compressor-raw-qc-Dispatcher.yaml
```diff
@@ -16,6 +16,11 @@ bind:
     transport: shmem
     addressing: ipc
     rateLogging: "{{ fmq_rate_logging }}"
+  - name: from_Dispatcher_to_dpl-output-proxy
+    type: push
+    transport: shmem
+    addressing: ipc
+    rateLogging: "{{ fmq_rate_logging }}"
```
knopers8, line 23: This and the other changes to tasks for `phos-compressor-raw-qc` are already there in `master`. I don't understand why github does not complain about merge conflicts to resolve.

scripts/phos-compressor-raw-qcmnt3.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
```
knopers8, line 1: Templates generated by this script are missing in the PR.

knopers8 (changes requested):
Thank you for the templates and the doc. There are minor things to be fixed/added.

Please also add the three new scripts to `scripts/generate-all-dpl-workflows.sh`

scripts/phos-compressor-raw-qcmnt3.sh
Barthelemy: correct, completely missed htis one.

tasks/phos-compressor-raw-qc-Dispatcher.yaml
Barthelemy: I will regenerate any ways, let's see if it stays.

scripts/phos-compressor-raw-qcmnt3.sh
Barthelemy: yep, forgot to push the commit. Regenerating any ways.

tasks/phos-compressor-raw-qc-Dispatcher.yaml
Barthelemy: the mystery stays: why does it mark it as new lines ?

knopers8: Could you rebase? For sure you don't have changes from the latest master in your branch.

Barthelemy: yep, that's it

knopers8 (approved):

knopers8:
@miltalex @teo Could you please bring this to the FLP v0.19.0 branch?

miltalex:
@knopers8 Yes I will do it now

PR #87, opened 2021-07-06
FT0: first version of workflow, raw->digit convertation and QC

afurs:


afurs:
Hello @teo ,
could you take a look to this PR please?

teo:
Looks ok but I kindly ask @knopers8 to confirm

afurs:
@teo thanks!
BTW, looks like checks are failed. Is it ok?

scripts/etc/ft0-digits-qc.json
```diff
@@ -0,0 +1,64 @@
...
+          "taskParameters": {
+            "ChannelIDs": "204,205"
+          },
+          "location": "local",
+          "localMachines": [
+            "localhost"
+          ],
+          "remoteMachine": "localhost",
+          "remotePort": "30132",
+          "mergingMode": "entire"
```
knopers8, line 46: the workflow you generated does not use mergers, so this is not needed (but it does not hurt). out of curiosity, why did you put this? as far as i remember FT0 uses one FLP, so there is no need for merging anyway.

knopers8 (approved):
looks good to me

knopers8:
One thing i forgot: can you add the script to the list in `scripts/generate-all-dpl-workflows.sh`? this way we can regenerate all the templates if something in DPL changes.

i noticed your hack for the QC cycle length configuration, this a good catch, i wasn't aware we can't modify it when running with ECS. I will note it down in jira.

scripts/ft0-digits-qc.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+set -x; # debug mode
```
knopers8, line 2: and please comment the debug mode

afurs:
Hello @knopers8 ,

I've made some modifications in FT0's dpl-workflow(some options were removed) in recent commit, also I commented debug mode line in script.

knopers8 (approved):
thanks

scripts/etc/ft0-digits-qc.json
afurs: > the workflow you generated does not use mergers, so this is not needed (but it does not hurt). out of curiosity, why did you put this? as far as i remember FT0 uses one FLP, so there is no need for merging anyway.

I was making some tests with this QC config, and forgot to remove "merging" option.

PR #106, opened 2021-08-24
Regenerate workflow templates for FLP suite v0.24

knopers8:
@teo This is needed to fix the multinode QC workflows in the FLP suite v0.24. Could you please bring this commit also to the corresponding v0.24 branch?

@Barthelemy FYI
@JianLIUhep This should fix your multinode setup hopefully.


knopers8:
FLP suite v0.23 is also broken actually, if there is a need, I can regenerate workflows also for that release (I would not risk assuming they are completely the same).

PR #112, opened 2021-09-02
Fix QC remote per detector

miltalex:


workflows/readout-dataflow.yaml
```diff
@@ -26,6 +26,14 @@ defaults:
     description: "Enable/disable EPN"
     widget: checkBox
     panel: BasicConfiguration
+  qc_remote_workflow: !public # TODO: if false, DCS panel should be disabled
```
vascobarroso, line 29: @miltalex I guess the TODO should be adapted. What we would like is that if this var is disabled, then all per-detector remote QC workflows vars are disabled.

vascobarroso (approved):

workflows/readout-dataflow.yaml
```diff
@@ -924,6 +932,12 @@ roles:
           - name: dpl
             enabled: "{{ qc_remote_workflow != 'none' }}"
             include: "{{ qc_remote_workflow }}"
+          - name: echo task
```
knopers8, line 935: Could you please add a comment explaining what is it for? Without the context that we are aware of (until we forget), this looks very unclear.

miltalex: Of course, I will create a second PR

miltalex: https://github.com/AliceO2Group/ControlWorkflows/pull/115

PR #118, opened 2021-09-09
Add mid-raw-decoder workflow

dstocco:
The PR contains 5 commits. The first one introduces a function that replaces the local variables in the generated yaml file.
It seems that this part of the code is used everywhere, so it might be better to include it in the global helper.
I tested that the function works on both linux and Mac (which have different sed).
The second commit contains the workflow generator script.
The third commit contains the generated files. These will change on the next flp-suite release since I renamed a class in O2.
The other commits makes the new workflow public, replacing the mid-raw-parser which was just a test and will not be used anymore. I plan to remove the corresponding files in a future PR.
The commits are meant to be atomic. Please do not squash.


knopers8 (approved):
@dstocco Thank you, also for contributing with the helper function. I will try adopting it in other workflow scripts when I have a moment.
It looks good to me.

@vascobarroso I run Diego's script on our staging setup with FLP suite v0.27, it generates an updated version of the one in this PR. I will make a PR soon.

PR #119, opened 2021-09-10
Add remote workflow for EMCAL getting data from FLP and EPN

mfasDa:
- Raw QC: FLP
- Digits QC: EPN


mfasDa:
@knopers8 @teo Could you please have a look whether this looks reasonable?

The goal is to enable raw QC on the FLP. As EMCAL has 2 FLPs we need a remote workflow which can merge both  QC histograms from the FLP and the EPN.

knopers8:
@mfasDa I think you are missing the local workflow in such case (the one for FLPs). Please also include the scripts and QC config files in the PR, so we can regenerate the templates for each FLP suite release.

mfasDa:
@knopers8 I have created a new local workflow for the raw QC remote setup on the FLPs. I have outsourced it to #124. Is there a way to generate the workflow template for the remote workflow in a similar way also for the QC remote workflow as described here (https://github.com/AliceO2Group/ControlWorkflows#exporting-dpl-workflow-templates - with the --o2-control option). Putting the workflow together manually is a bit error prone.

knopers8:
@mfasDa Sorry, I do not understand your question. In general we do not invest too much time into the procedure of generating these templates, because we plan to move to Just-In-Time translated workflows, though it seems to be dragging a bit.

As I mentioned already before, please also include the scripts and QC config files in the PR, so we can regenerate the templates for each FLP suite release.

knopers8:
Also [here](https://github.com/AliceO2Group/ControlWorkflows/blob/9b6870b7eba1c416e33911b4f70aaa7c1fe0ba54/workflows/readout-dataflow.yaml#L75) you probably want to add your new local workflows, so they appear in the selection box. A bit [below](https://github.com/AliceO2Group/ControlWorkflows/blob/9b6870b7eba1c416e33911b4f70aaa7c1fe0ba54/workflows/readout-dataflow.yaml#L87) there is a similar list for the remote workflows.

mfasDa:
@knopers8 I found under scripts what I was looking for (emc-qcmn-epn.sh). I will use that as template for the other workflows.

mfasDa:
Closing due to branch restructuring based on EMCAL JIRA ticket IDs.

PR #123, opened 2021-09-14
reduce TOF QC sampling on FLP

noferini:


noferini:
I would reduce the sampling in FLP TOF QC.
If I am not wrong this should be enough.

PR #124, opened 2021-09-14
Add raw QC workflow for EMCAL with multi-node setup

mfasDa:
- Rasks for raw QC
- Push to merger alio2-cr1-qc02


mfasDa:
@knopers8 @teo Could you please review?

knopers8 (changes requested):
Please keep the same naming convention for the local and remote parts. In this case the local part should be called `emc-qcmn-flp-local`, following how you called the remote part.

Please also include the scripts and QC config files in the PR, so we can regenerate the templates for each FLP suite release.

mfasDa:
Thanks @knopers8! I have prepared scripts and QC configs and will add them. I am fine with the naming convention, it makes sense and will simplify life - I will change accordingly.

Closing due to branch restructuring based on EMCAL JIRA ticket IDs.

PR #129, opened 2021-09-16
[EMCAL-718,EMCAL-720,EMCAL-721,EMCAL-722] Control workflows for EMCAL in multi-node setup

mfasDa:
- [EMCAL-720] Control workflow for raw QC on FLP
- [EMCAL-721] Remote workflow for raw QC from FLP
- [EMCAL-722] Remote workflow for raw QC from FLP and digits QC from EPN


mfasDa:
@knopers8 @teo Please review! Supersedes #119 and #124

mfasDa:
Control workflows also added to the EMCAL workflow menues

scripts/etc/emc-qcmn-flpepn.json
```diff
@@ -0,0 +1,215 @@
...
+                "dataSource": {
+                  "type": "dataSamplingPolicy",
+                  "name": "emcrawdata"
+                },
+                "location": "local",
+                "localMachines": [
+                    "localhost"
+                ],
+                "remoteMachine": "alio2-cr1-qc02",
+                "remotePort": "47702",
```
knopers8, line 42: You used the same port for RawTask and DigitsQC. This will probably work, but only because the RawTask will run controlled by AliECS, so the port will be overwritten by AliECS.
The EPN part (DigitsQC) has a statically set port and it actually will be 47702.
You could consider using a different one just for the sake of clarity, because these tasks do not use the same Merger process.

scripts/etc/emc-qcmn-flpepn.json
```diff
@@ -0,0 +1,215 @@
...
+                "mergingMode": "delta"
+            },
+            "DigitsQC": {
+                "active": "true",
+                "className": "o2::quality_control_modules::emcal::DigitsQcTask",
+                "moduleName": "QcEMCAL",
+                "detectorName": "EMC",
+                "cycleDurationSeconds": "10",
```
knopers8, line 50: Please also notice that for now, the cycle durations are taken from the workflow template generation time. This is due to how timers are configured in DPL. Once we have JIT translation for those templates, the values from consul will actually take precedence.

scripts/etc/emc-qcmn-flpepn.json
```diff
@@ -0,0 +1,215 @@
...
+                {
+                    "condition": "random",
+                    "fraction": "0.1",
+                    "seed": "1248"
+                }
+            ],
+            "blocking": "false"
+        }
+    ]
+}
```
knopers8, line 215: If you do not put any specific `machines` in sampling policies, they will be spawned for each workflow which uses this file. If you want to use the same one for FLPs and EPNs, I suggest that use some keywords like "flp" and "epn" in machines (also in "localMachines" in tasks) and then run the local QC workflows with `--local --host flp` ro `--local --host epn`.

knopers8 (changes requested):
Please add the scripts to `generate-all-dpl-workflows.sh` so we can regenerate the templates after each FLP suite release in one go.

scripts/etc/emc-qcmn-flpepn.json
mfasDa: Can we freely decide on the port, i.e. if we choose 47703 will this work? (firewall open?)

scripts/etc/emc-qcmn-flpepn.json
mfasDa: This needs some clarification: Do we need to list under machines the full hostname (i.e. alio2-cr1-flp146 and similar for EPN nodes) or is the tag flp / epn sufficient as machine name? Ideally we would restrict to 1 configuration. For the current tests we have created different configurations for raw QC on FLP and digits QC on EPN

knopers8: From https://github.com/AliceO2Group/QualityControl/blob/master/doc/Advanced.md#multi-node-setups :
> One also may choose the merging mode - delta is the default and recommended (tasks are reset after each cycle, so they send only updates), but if it is not feasible, Mergers may expect entire objects - tasks are not reset, they always send entire objects and the latest versions are combined in Mergers. With the delta mode, one can cheat by specifying just one local machine name and using only that one during execution. This is not possible with entire mode, because then Mergers need identifiable data sources to merge objects correctly.

scripts/etc/emc-qcmn-flpepn.json
knopers8: From readme:
> Please use the port number between 47700 and 47799.

I did not find other usages of 47703 in the repository, so I think you should be safe to use it.

scripts/etc/emc-qcmn-flpepn.json
mfasDa: The question was not related to the merging mode but concerning the local nodes, particular with respect to running on the EPN: It is not a problem to list the two FLP nodes for the tasks that run on the FLP, but on the EPN EMCAL runs on the dynamic allocation where it will be decided on the fly which hosts will be used - do we have to put there ALL EPN nodes as local machines?

knopers8: Let me repeat the most crucial sentence there:

> With the delta mode, one can cheat by specifying just one local machine name and using only that one during execution.

You can use just e.g. "epn" or "localhost" there, and invoke QC with `--local --host epn` or `--local --host localhost`. That is, if you use the default merging mode, as indicated in the doc.

mfasDa: I see, so what we add on the command line for `--host` must match `localMachines` in the config.json. OK, then we will go consistently for `flp` and `epn`. I guess this is not only needed in the data sampling policy but also in the task configuration, right?

knopers8: yes. if i remember correctly, leaving "machines" and "localMachines" empty will match any `--host` in both cases.

mfasDa: OK, thanks for the confirmation. Fixed in the latest version.

mfasDa:
@knopers8 I have fixed all requests and restructured the commits. Please have a look whether this can be merged now!

knopers8 (approved):

PR #131, opened 2021-09-21
[EMCAL-718] Set Cycle duration to 60 seconds

mfasDa:


mfasDa:
@knopers8 Setting cycle duration to 60 seconds as discussed on the FLP-EMCAL channel.

knopers8 (approved):

PR #132, opened 2021-09-21
[EMCAL-718, EMCAL-722] Fix cycle duration for digits merger in combined remote workflow

mfasDa:


mfasDa:
@knopers8 Also concerning the cycle duration.

knopers8 (approved):

PR #135, opened 2021-09-22
Add PDP panel

vascobarroso:
Introduce changes as described in https://alice.its.cern.ch/jira/browse/O2-2591.


vascobarroso:
@teo @davidrohr this should be merged once the corresponding changes are done in the AliECS/ODC interface

PR #144, opened 2021-10-01
Add HMPID raw to pedestals workflow

kostorr:
Closes ODC-137


knopers8 (changes requested):
Please add the script to `generate-all-dpl-workflow.sh`, the rest looks ok

knopers8 (approved):

PR #146, opened 2021-10-04
[WIP] MFT full sync multinode QC workflow

knopers8:


knopers8:
We did not test it yet, but since we have little time, we could already merge it to master. In the worst case we fix it, but we will not destroy anything else by adding this workflow.

PR #147, opened 2021-10-05
[EMCAL-718,EMCAL-725]Adding remote workflow for running raw QC on the…

mfasDa:
… EPN

In case both raw and digits QC are running on the EPN
a separate remote workflow is needed ot fix different
ports for the raw and digits proxy as the raw proxy
is no longer handled by ecs but by odc.


mfasDa:
@teo @knopers8 This workflow is needed in case both QC tasks run on the EPN (plan for the pilot beam). We have been testing this the last week at P2, it should be safe to merge.

PR #149, opened 2021-10-06
CPV pedestal calibration workflow + small fixes

teo:
@sevdokim


sevdokim:
Hi Teo! Yes, it seems that everything is runable in aliecs so please merge it.
Cheers,
Sergey

PR #155, opened 2021-10-09
CTP integration

miltalex:


vascobarroso (approved):

vascobarroso (approved):

vascobarroso:
@miltalex there were some lines in this PR that looked like a leftover from an older version of this workflow. I have removed them, for me it looks fine now. 

One small question: shouldn't we include a toggle button in the GUI to enable or disable the sending of commands to the TRG gw ? Like it is now, it looks to me that we don't have a choice, we will always send (which might forbids us to continue to use the old manual LTU operations, not sure if we are ready for that).

vascobarroso:
@miltalex many thanks! Something like we have for the DCS would be perfect.

workflows/readout-dataflow.yaml
```diff
@@ -1054,3 +1105,33 @@ roles:
...
+        call:
+          func: ctp.RunStop()
+          trigger: before_STOP_ACTIVITY
+          await: after_STOP_ACTIVITY
+          timeout: 5s
+          critical: true
+      - name: unload
+        call:
+          func: ctp.RunUnload()
+          trigger: after_STOP_ACTIVITY
```
teo, line 1135: This is the `STOP_ACTIVITY` transition sequence as it happens:

- `before_STOP_ACTIVITY`
  - call `RunStop`
- `leave_RUNNING`
- `STOP_ACTIVITY` transition
- `enter_CONFIGURED`
- `after_STOP_ACTIVITY`
  - call `RunUnload`
  - await `RunStop`
  - await `RunUnload`

As you can see it might happen that `RunUnload` starts before `RunStop` is done. Is this acceptable?
If not, `RunStop` should await an earlier target moment, probably `before_STOP_ACTIVITY`.
Within the same trigger moment, all call start triggers are processed before call await triggers.

teo (approved):

PR #161, opened 2021-10-14
regenerate DPL workflows for FLP Suite v0.32.0

vascobarroso:


vascobarroso:
@knopers8 does this look ok ?

PR #167, opened 2021-10-18
fix in tof json script and add postprocessing

noferini:


Barthelemy:
I think that it always fails. @teo  might be able to say more

noferini:
Thanks

PR #189, opened 2021-11-01
Regenerate workflows for 0.34.1

vascobarroso:


vascobarroso:
@knopers8  we have picked up a brand new O2 this morning to reduce the number of monitoring metrics sent by DPL. This PR is the outcome of the regeneration, looks ok right ?

knopers8 (approved):
I think it is OK, I found some corresponding code:
```
std::string composeIPCName(std::string const& prefix, std::string const& hostname, short port)
{
  if (prefix == "@") {
    return fmt::format("ipc://@{}_{},transport=shmem", hostname, port);
  }
```
It seems that the fairmq ipc prefix will be now deduced automatically.
Perhaps @ktf could give us an additional confirmation that this is expected.

PR #191, opened 2021-11-06
regenerate DPL workflows for FLP Suite v0.35.0

vascobarroso:


vascobarroso:
@knopers8 there was an error when regenerating the workflows: 

```
[INFO] ...created.
[INFO] Creating a task dump for 'MERGER-QcMFTClusterTask1l-0'.
[INFO] ...created.
[INFO] Creating a task dump for 'MERGER-QcMFTReadoutTask1l-0'.
[INFO] ...created.
[INFO] Creating a task dump for 'QC-CHECK-RUNNER-QcMFTClusterCheck'.
[INFO] ...created.
[INFO] Creating a task dump for 'QC-CHECK-RUNNER-QcMFTReadoutCheck'.
[INFO] ...created.
[INFO] Creating a task dump for 'internal-dpl-injected-dummy-sink'.
[INFO] ...created.
[ERROR] Error: cannot open file /root/ControlWorkflows/scripts/etc/mid_electronics_delay.txt
[INFO] Dumping the workflow configuration for AliECS.
[INFO] Creating directories 'workflows' and 'tasks'.
[INFO] ... created.
[INFO] Creating a workflow dump 'mid-raw-decoder'.
[INFO] This topology will connect to the channel 'readout-proxy', which is most likely bound outside. Please make sure it is declared in the global channel space under the name 'readout-proxy-{{ it }}' in the mother workflow or another subworkflow.
[INFO] This topology will bind a dangling channel 'downstream' and declare it in the global channel space under the name 'downstream-{{ it }}'. Please make sure that another device connects to this channel elsewhere. Also, don't mind seeing the message twice, it will be addressed in future releases.
```

Could you check if it is relevant or if it can be ignored ? Thanks!

knopers8:
@vascobarroso I suppose it should be fine if the `mid-raw-decoder` workflow template was still created and stayed the same. @dstocco could you confirm this error is harmless? (see above)

dstocco:
Well, this part was meant to be temporary. The workflow needs 2 configuration (txt) files. But if it does not find them, it should use default ones, which are ok. The error comes from the fact that those files are not in the local etc dir. This is not an issue. But the point is that the file are currently not in the "final" directory on the FLP, which was chosen as `/tmp` since it is the only directory accessible by both `flp` and `mid` users. And `/tmp` is erased from time to time.
So I will have to fix this eventually, but maybe you can go on now and I'll fix it for the next FLP suite.

teo (approved):

PR #193, opened 2021-11-12
Remove custom configuration files from MID raw decoder

dstocco:
This PR should fix the error seen in this PR:
https://github.com/AliceO2Group/ControlWorkflows/pull/191
The main commit is the first one. The second only generates the corresponding files (although it might be not needed since they are regenerated at each FLP suite release).
So the second commit could be removed, if needed.
Please do not squash.


dstocco:
Not sure to understand the error message: it is complaining that the (arbitrary) name that I chose for my branch is not known to git...
Sounds like a CI bug. Please let me know if some action is needed from my side...

graduta:
> Not sure to understand the error message: it is complaining that the (arbitrary) name that I chose for my branch is not known to git... Sounds like a CI bug. Please let me know if some action is needed from my side...

Hi @dstocco . Indeed this is a bug in the CI as forked repositories were not taken into consideration for the checks.
I checked your PR myself for changes that might affect the GUI and there is nothing wrong with it.

We will update the CI Checks soon to work for forked repositories as well

knopers8 (approved):

PR #197, opened 2021-11-15
Set --discovery-endpoint on stfsender only if DD scheduler is enabled

vascobarroso:


vascobarroso:
@ironMann, this seems to do the trick for DD 1.0.14 (i.e. I could start a run with the CI config). 
Still, `StfBuilder` also complains with the following messages (doesn't seem to block the start of a run): 

```
E | 18:24:32.672 | ansible-ci-cs8-1 | FLP | datadist/stfbuilder | 1 |   | Consul kv set error. what=Protocol "no-op" not supported or disabled in libcurl (1)
W | 18:24:37.676 | ansible-ci-cs8-1 | FLP | datadist/stfbuilder | 1 |   | Consul kv param retrieve error. what=Protocol "no-op" not supported or disabled in libcurl (1)

```
Do we also need to change something there ?

vascobarroso:
@ironMann ok, we will merge this and wait for your new DD release.

ironMann:
Hi @vascobarroso , OK. The next version of DD will require the latest FairMQ. Is it included in the FLPSuite? O2 is already building it on EPNs, so we should keep this in sync.

vascobarroso:
@ironMann the current FLP Suite version installed at P2 (0.36.0) still uses FairMQ v1.4.42. The next FLP Suite (0.37.0), to be released on Thursday and deployed next Monday, will include the latest bumps from `alidist@master`, including FairMQ v1.4.43 which was bumped last Friday. 

By the way, the change done in this PR was enough to get back to a green CI. Still, it would probably be a good idea to get rid of the StfBuilder errors.

PR #198, opened 2021-11-16
[EMCAL-718] Change QCDB URL to new relocator

mfasDa:
- Use ali-qcdb.cern.ch:8083 as QCDB URL in QC configs
- Enable payload size cheker for event-based histogram


knopers8 (approved):

scripts/etc/emc-qcmn-epn.json
```diff
@@ -3,7 +3,7 @@
         "config": {
             "database": {
                 "implementation": "CCDB",
-                "host": "alio2-cr1-hv-qcdb1.cern.ch:8083",
+                "host": "ali-qcdb.cern.ch:8083",
```
knopers8, line 6: FYI, here it does not matter too much, because the QC will anyway ask consul for the config file to be used and the QCDB host name does not have influence on the workflow structure.

mfasDa: Yes, this is mostly for reference. I hope a solution will be found also for the EPN to use the configs in consul for the common workflow, I would prefer to have all configs in one place and not mess around with duplications of the same file in 2 different repositories  + a key/value database.

knopers8: Yes, that will be done.

PR #209, opened 2021-12-09
Switch from O2DataProcessing to O2DPG

davidrohr:
The repositories have been merged. Thus this PR changes some of the field names / descriptions to be more clear.
We will test the new repository tomorrow, Friday 10th at 2pm. Please do not merge before.
@teo : Note that I have left the internal field names, e.g. `pdp_o2_data_processing_path` as they are, since I assume otherwise the core would need to be changed.

### DIFF
--- a/workflows/readout-dataflow.yaml
+++ b/workflows/readout-dataflow.yaml
@@ -871,16 +871,16 @@ defaults:
     value: "v0.20"
     type: string
     label: "O2 Data Processing Hash"
-    description: "Commit hash or tag of O2DataProcessing to use"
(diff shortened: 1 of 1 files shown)


davidrohr:
@teo @vascobarroso : We switched the setup on the EPNs to use O2DPG from now, and this was tested and works.
Can you merge this PR, to update the texts in the ECS GUI.

vascobarroso (approved):

vascobarroso:
@davidrohr done, it will be available from FLP Suite v0.40.0, to be deployed at P2 next Monday morning.

PR #211, opened 2021-12-15
Reformat multiline roc-config call

teo:


teo:
@kostorr ↑ for your information
@knopers8 not urgent but could you please ensure that the DPL generator outputs the `_plain_cmdline` under a YAML multiline block (`>-`) same as `_module_cmdline`?

knopers8:
@teo OK, will do

PR #212, opened 2021-12-17
TPC QC remote workflow has now mergers for all tasks

knopers8:
Waiting for confirmation from TPC that it works.


knopers8:
TPC confirmed it works.

vascobarroso (approved):
@knopers8 looks good and therefore approving but I didn't really test. Has this been tried at P2 ?

knopers8:
Yes

PR #217, opened 2022-01-25
A workaround TPC QC remote workflow until OCTRL-591 is resolved

knopers8:


knopers8:
Robert requested to have the workaround workflow in the next FLP suite, so they can participate in global runs.

vascobarroso (approved):

PR #225, opened 2022-01-28
Regenerate two updated TPC workflows

knopers8:


knopers8:
@vascobarroso @teo Could you please bring this one to flp-suite-v0.44 as well?

PR #226, opened 2022-01-29
Update PDP/EPN panel

davidrohr:


workflows/readout-dataflow.yaml
```diff
@@ -921,16 +947,29 @@ defaults:
...
+    type: number
+    label: "% of raw data to store (not working)"
+    description: "In case of a 'disk' TfBuilder mode, selects the fraction of raw data to be stored in percent"
+    widget: comboBox
+    panel: "EPN_Global_Processing"
+    values:
+      - "production/production.desc"
+      - "production/no-processing.desc"
+    index: 102
+    visibleif: ($$pdp_config_option === "Repository hash" || $$pdp_config_option === "Repository path") && ($$tfb_dd_mode === "disk" || $$tfb_dd_mode === "processing-disk") $$pdp_panel_mode === "Expert"
```
teo, line 964: It cannot be `type: number` with an enum of `values`. I assume the `type` is correct, but `values` needs to go.

teo (changes requested):

workflows/readout-dataflow.yaml
davidrohr: you are of course right, the values need to go, and perhaps also the widget should be adapted then. Thx for spotting!

teo:
@vascobarroso in the interest of having this ready for next week I'll merge to CW master as soon as @davidrohr sorts out a minor issue. Some indexes have been changed, so the **visual** aspect of the panel will need to be revalidated. This can be done now already, with any version of AliECS (though we'll need this week's upcoming tag to test the **functionality**).

davidrohr:
@teo : I have fixed the issue you mentioned. The change of some of the indexes was actually done to make it visually more clear (and in fact I didn't reorder much, but I grouped things belonging together in index ranges with gaps, such that one can add fields in between more easily in the future).

PR #231, opened 2022-02-04
Update default O2DPG version

davidrohr:


davidrohr:
@teo @vascobarroso @pippohub : Any change to also get this change in for the FLPSuite update on Monday? It just changes the default version of the O2DPG repo used to generate workflows.

teo (approved):

PR #234, opened 2022-02-10
[OCTRL-564] CCDB integration

knopers8:
We need first https://github.com/AliceO2Group/Control/pull/322 in the same FLP suite release.


workflows/readout-dataflow.yaml
```diff
@@ -1501,3 +1509,19 @@ roles:
           trigger: DESTROY-100
           timeout: 5s
           critical: false
+  - name: ccdb
+    enabled: "{{ccdb_enabled == 'true'}}"
+    defaults:
```
teo, line 1514: Useless line is useless, otherwise all good!

teo (changes requested):

workflows/readout-dataflow.yaml
knopers8: Indeed.

teo (approved):

PR #240, opened 2022-02-15
FIT: global renaming for QC tasks

afurs:
and calib-offset modifications for FT0/FV0


knopers8 (approved):
When the changes are in the new FLP suite deployed at P2, please remember to update the QC config files in consul accordingly.

PR #244, opened 2022-02-16
[EMCAL-524] Fix renamed binding

mfasDa:
For consistency the binding emcal-digits
was also renamed to emcal-cells


mfasDa:
@knopers8 Sorry, forgot to adapt the binding for the cell input. The binding for the trigger input was not changed.

PR #251, opened 2022-03-01
[OCTRL-564] Enable GRP object upload by default

knopers8:


vascobarroso:
@shahor02 this PR will enable by default the creation of the GRP object in CCDB by AliECS for each run. Should we merge this to have it for the deployment of next Monday at P2 ? Or do you still want to do some extra tests ?

In addition, the version of next Monday will also include the run type.

vascobarroso:
@shahor02 could you comment on this ? Should we enable this by default at P2 or do you want to do some extra tests ?

vascobarroso:
Ok thanks!

vascobarroso:
@knopers8 in fact, now that I think about it, should we really enable this at the level of the workflow ? AFAIK, we only want to enable this at P2, not on the lab/CI setups. If so, we should set it to true in Consul.

knopers8:
@vascobarroso indeed, you are right. I just added the value in `o2/runtime/aliecs/vars/ccdb_enabled`, it seems to work.

PR #253, opened 2022-03-02
Hmpid new raw to pedestals

fapfap69:


scripts/hmpid-raw-to-pedestals_1.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+# set -x;
+set -e;
+set -u;
+
+WF_NAME=hmpid-raw-to-pedestals
```
knopers8, line 7: If you need a new workflow, please give it a different name, so it does not overwrite the one which already exists, add it to the list of automatically generated workflows in `scripts/generate-all-dpl-workflows.sh` and to the list in `readout-dataflow.yaml` for HMPID.
However, it seems to me that you only want to change the existing workflow. In such case, I would recommend that you modfiy the existing script.

workflows/hmpid-raw-to-pedestals1.yaml
```diff
@@ -0,0 +1,69 @@
+name: hmpid-raw-to-pedestals
```
knopers8, line 1: This would be needed only if you need a new workflow.

knopers8 (changes requested):
Please see my comments.
I noticed you brought also commits from one of `flp-suite` branches which are not present in `master`. Please remove them from this PR.

PR #257, opened 2022-03-04
Remove ccdb-url from CPV workflows, as it does not exist anymore

knopers8:


knopers8:
@sevdokim when regenerating workflow templates for the latest FLP suite release, we discovered the `--ccdb-url` does not exist anymore, so we remove it. Please let us know if it should be replaced with something else.

vascobarroso (approved):

PR #260, opened 2022-03-08
[OCTRL-161] Add a description field to the main workflow templates

claireguyot:


vascobarroso:
@knopers8 could you provide a more detailed description of the 2 QC workflows for @claireguyot to add ?

workflows/qc-postprocessing.yaml
```diff
@@ -1,4 +1,5 @@
 name: !public qc-postprocessing
+description: !public "Qc-postprocessing worflow template"
```
knopers8, line 2: A workflow template for running QC post-processing as OCC State Machine (obsolete)

workflows/qc-server-workflow.yaml
```diff
@@ -1,4 +1,5 @@
 name: !public qc-server-workflow
+description: !public "Qc-server-workflow worflow template"
```
knopers8, line 2: A workflow template for running standalone QC workflows on QC servers.

workflows/qc-postprocessing.yaml
vascobarroso: @knopers8 by obsolete does it mean it we can remove it ?

knopers8: It seems that `qc-server-workflow` does not cause problems, so we can indeed remove it. I will make a PR.

workflows/qc-server-workflow.yaml
vascobarroso: @claireguyot could you update this description with the suggestion from @knopers8 ?

workflows/qc-postprocessing.yaml
vascobarroso: Ok thanks. @claireguyot I guess you will need to rebase and drop these changes.

workflows/qc-server-workflow.yaml
claireguyot: @vascobarroso The changes have been added!

vascobarroso (approved):

PR #263, opened 2022-03-10
Pass O2_PARTITION to Readout and other tasks

teo:


tasks/o2-roc-cleanup.yaml
```diff
@@ -13,7 +13,9 @@ wants:
 command:
   log: "{{ log_task_output }}"
   shell: true
-  env: ["O2_DETECTOR={{ detector }}"]
+  env:
+    - O2_DETECTOR={{ detector }}
+    - O2_PARTITION={{ environment_id }}
```
teo, line 18: @knopers8 can you please do this ↑ in the DPL generator?

knopers8: i will

vascobarroso (approved):

ironMann:
Sorry @vascobarroso , too much stuff happening today. I'm not sure about how that mechanism works. Maybe I'm using infologger in an incompatible way. I set the run and partition values manually. 

If you add this value to `StfBuilder --discovery-partition=`  cmdline parameter, then I can pick it up on start as well.

vascobarroso:
@sy-c could you provide some guidance here ? My intuition tells me that the most flexible would be to use the env vars (since it would allow us to set/change fields at the level of the workflow definition without any code change) but I don't have the full picture.

sy-c:
O2_PARTITION and O2_DETECTOR environment variables is the way to go, they are picked up directly by infoLogger. Launched processes don't have to take care of them.

vascobarroso:
@ironMann but do you understand why the env var O2_PARTITION has no effect on FLPs ? According to @sy-c explanation, they should be picked up automatically by the infologger library. Do you by any chance explicitly unset some vars at startup ?

PR #264, opened 2022-03-10
regenerate DPL workflows for FLP Suite v0.49.0

vascobarroso:


knopers8 (approved):
New MFT task names as anticipated, all seems ok.

PR #265, opened 2022-03-14
Cpv noise calibration

sevdokim:
Hello!
CPV noise calibration workflow is added to repo. Also DCS CCDB populator is added to cpv-pedestal-calib-qc workflow. Also localhost:8084 is added as condition backend to cpv-physics workflow.


sevdokim:
Hi! Would it be possible to include the changes in this PR into FLP suite 0.50 (next one which will appear this Thursday)?
The changes are essential for cpv noise and physics runs.

scripts/etc/cpv-qc-compressor.json
```diff
@@ -28,7 +28,7 @@
         "className": "o2::quality_control_modules::cpv::PedestalTask",
         "moduleName": "QcCPV",
         "detectorName": "CPV",
-        "cycleDurationSeconds": "100",
+        "cycleDurationSeconds": "10",
```
knopers8, line 31: Do you really need 10 seconds? This will actually have an effect on the cycle duration.

knopers8 (approved):

PR #266, opened 2022-03-15
Fix the MFT QC digits task inputs

knopers8:


knopers8:
@tomas-herman FYI
@teo could you please bring it to the latest flp-suite branch?

teo (approved):

PR #277, opened 2022-03-28
add ccdb server key value

Barthelemy:


vascobarroso:
Close in favor of #281

PR #278, opened 2022-03-28
Fix MFT FLP workflow

vascobarroso:


vascobarroso:
Close in favor of #281

PR #279, opened 2022-03-29
Specify FLP/DISTSUBTIMEFRAME/0 rather than wildcarded version

ktf:


ktf:
The only FLP/DISTSUBTIMEFRAME which can come from datadistribution is FLP/DISTSUBTIMEFRAME/0, and it's actually an error to have the wildcard, because it will break the handling of race conditions with optionals.

vascobarroso:
Closed in favor of #280

PR #284, opened 2022-04-05
[OCTRL-551] Kafka integration

knopers8:


knopers8:
@vascobarroso I created a button to be able to debug more easily, but we can hide the setting already in consul, if you prefer.

vascobarroso (approved):

PR #286, opened 2022-04-07
daq-qcmn: detector variable instead of TST

Barthelemy:
Allows to follow these instructions: https://github.com/AliceO2Group/QualityControl/blob/master/doc/Advanced.md#switch-detector-in-the-workflow-readout-dataflow

### DIFF
--- a/scripts/qcmn-daq.sh
+++ b/scripts/qcmn-daq.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# set -x;
+set -x;
 set -e;
 set -u;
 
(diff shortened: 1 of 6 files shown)


scripts/qcmn-daq.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# set -x;
+set -x;
```
knopers8, line 3: Please comment it back, so it does not pollute the output when regenerating all the workflows

scripts/qcmn-daq.sh
```diff
@@ -43,8 +48,14 @@ sed -i /defaults:/\ a\\\ \\\ "${QC_CONFIG_PARAM}":\ \""${ESCAPED_QC_FINAL_CONFIG
 ESCAPED_QC_GEN_CONFIG_PATH=$(printf '%s\n' "$QC_GEN_CONFIG_PATH" | sed -e 's/[]\/$*.^[]/\\&/g');
 sed -i "s/""${ESCAPED_QC_GEN_CONFIG_PATH}""/{{ ""${QC_CONFIG_PARAM}"" }}/g" workflows/${WF_NAME}.yaml tasks/${WF_NAME}-*
 
+# replace TST with a parameter
+#sed -i "s/TST/{{ detector }}/g" workflows/${WF_NAME}.yaml tasks/${WF_NAME}-*
+
+# add a default value for detector
+#sed -i /defaults:/\ a\\\ \\\ "detector: TST" workflows/${WF_NAME}.yaml
+
```
knopers8, line 56: please remove if not needed

knopers8 (changes requested):
thanks, i would just ask for two small fixes

scripts/qcmn-daq.sh
Barthelemy: absolutely

knopers8 (approved):

PR #287, opened 2022-04-07
Tpc remote workflows

rmunzer:
Adding QC Nde workflows which includes post-processing.


knopers8 (changes requested):
please include the new workflow generation scripts in `scripts/generate-all-dpl-workflows.sh`

rmunzer:
Hi @knopers8,
i just had a look. The scripts should be all in generate-all-dpl-workflows.sh

knopers8 (approved):
Perhaps I missed it before, sorry.

PR #294, opened 2022-04-15
ITS QC workflow on EPN

JianLIUhep:


JianLIUhep:
Would it be possible to include this workflow in FLP suite v0.54? Thank you.

knopers8 (approved):

PR #302, opened 2022-04-28
ft0 digits pipe: increased to 7

afurs:


knopers8 (approved):
Out of curiousity, have you tested it? We saw some problems with TOF compressors if there were too many of them in parallel, AliECS could not start the workflow in such case.

afurs:
Hi @knopers8 yes I tested it.

PR #304, opened 2022-04-29
Enable bookkeeping plugin by default

vascobarroso:


vascobarroso:
@teo @claireguyot given that we ship bookkeeping by default as part of the FLP Suite, any objection to enabling it by default ?

teo (approved):

PR #305, opened 2022-04-30
PHS qc remote workflow: Raw&Cluster tasks on EPNs

sevdokim:
Hello!

This is qc remote workflow for PHS to fetch RawTask and ClusterTask results from EPN when no processing is ongoing on FLP and everything happens at EPN. 
`"localMachines": ["localhost"]` is used in runtime configuration in consul as EPN requires it to be localhost now.


scripts/etc/phos-raw-clusters-epn.json
```diff
@@ -0,0 +1,100 @@
...
+  "dataSamplingPolicies": [
+    {
+      "id": "phosraw",
+      "active": "true",
+      "machines": ["epn"],
+      "query": "rawerr:PHS/RAWHWERRORS;cells:PHS/CELLS;cellstr:PHS/CELLTRIGREC",
+      "samplingConditions": [
+        {
+          "condition": "random",
+          "fraction": "1.",
```
knopers8, line 79: In the templates any value is fine of course, but if you actually intend to use the workflow with 100% data, please connect directly to the data source to get rid of unnecessary copies, as described [here](https://github.com/AliceO2Group/QualityControl/blob/master/doc/ModulesDevelopment.md#bypassing-the-data-sampling)

knopers8 (approved):

scripts/etc/phos-raw-clusters-epn.json
sevdokim: Hi @knopers8 
yes, you are right. For cpv I connect it derectly, here in phos some leftover remained. Thanks for pointing this out!

peressounko: Dear @knopers8, could you clarify one point related to direct/sampling data access: do I understand correctly that in "direct" case crush in QC part will also kill datataking flow, while in case of sampling it is kind of fuse/detach and crush of QC will not block datataking?

knopers8: That depends. In the current state of the DPL framework, a crash in any device (like QC task) in a workflow will crash the whole workflow. We want to improve it, but we are not there yet.

However, if we sample on EPN, but send data to a QC server (so technically we have two workflows), then a crash in a QC task should not cause the EPN workflow to crash as well.

PR #316, opened 2022-05-02
MCH: changed remotePort for errors task

aferrero2707:
The remote port number is changed from 47790 (already in use for the digits task) to 47791, which is currently unused.


knopers8 (approved):

knopers8:
please keep in mind, that without regenerating the workflow templates, the issue will be fixed only once we regenerate them all for the next release.

PR #318, opened 2022-05-03
Use alternative SHM segment for DD tasks

ironMann:
This is separate DD from other DPL managed shm segments. With such configuration, DD tasks don't have to be configured with the same SHM-size option as the rest of FLP workflows.
https://alice.its.cern.ch/jira/browse/R3C-854

### DIFF
--- a/tasks/stfbuilder.yaml
+++ b/tasks/stfbuilder.yaml
@@ -38,6 +38,7 @@ command:
   value: "{{ len(modulepath)>0 ? _module_cmdline : _plain_cmdline }}"
   arguments:
     - "--session=default"
+    - "--shm-segment-id=2"
     - "--transport=shmem"
     - "--monitoring-backend='{{ monitoring_dd_url }}'"
     - "--discovery-partition={{ environment_id }}"
--- a/tasks/stfbuilder.yaml
+++ b/tasks/stfbuilder.yaml
@@ -39,6 +39,7 @@ command:
   arguments:
     - "--session=default"
     - "--shm-segment-id=2"
+    - "--shm-segment-size=134217728"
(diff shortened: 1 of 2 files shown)


vascobarroso:
@ironMann should this be part of the next FLP Suite release ? Or is it still under discussion/validation ?

vascobarroso:
@ironMann I see. Could you expand a bit on "all FLP-workflows which convert raw data (TOF, CPV, ?) require this option for proper operation" ? AFAIK, we have been running these types of workflows since months.

ironMann:
The issue here is that all FMQ tasks in the chain must be configured with the same shm-size parameter. Otherwise, the first task that starts will create the SHM of default size. I don't know how this race is resolved exactly. Perhaps the tasks that are requiring 10 GB get to be limited to 2 GB (default value if not provided).  
Without this change, DD tasks would need to be configured with the same shm-size option as all QC and DPL processes on FLP. But since DD does not use the FMQ-DPL-SHM segment (data is in the readout SHM), we can add this option in order to avoid configuring the same shm-size for DD tasks.

@rbx @davidrohr  can perhaps comment on how the shm-size is resolved if all tasks do not have the same shm-size option. I know we had the same situation with TfBuilder and DPL-processing on EPN, and we resolved it in this way.

davidrohr:
Whoever is first creates the segment with size it has configured.
All other processes will use the segment as is, even if the size is lower than configured.

vascobarroso:
@ironMann thanks for the detailed explanations. Could you then target this for master instead of 0.56.0 ? I will then cherry pick it into 0.57.0.

ironMann:
Sure, moving to #325

PR #323, opened 2022-05-05
ITS QC workflow for FEE on FLP and tasks on EPN

JianLIUhep:
Hi @vascobarroso @knopers8, 
Could you please have a look this PR and (if possible) also merge it to the default revision which is being used at P2 by shifters? Then we can run ITS QC FEE on FLP again, which is very important to understand the detector performance. 

Thanks a lot.


knopers8 (approved):

knopers8:
It should be also on `flp-suite-v0.56.0`. You might need to wait a few minutes or refresh the repository.

JianLIUhep:
Thank you. It works.

PR #328, opened 2022-05-10
Move QC remote machine attribute and the corresponding fairmq cleaner to the main wf

knopers8:


knopers8:
@Barthelemy @kostorr This gets rid of the usual template processing on qc remote workflows. We will have to add corresponding `det_qc_remote_machine` values to consul or as defaults to `readout-dataflow`, but this is fine i guess.
To be tested at P2 yet, on staging it was fine.

Barthelemy:
Capture d’écran 2022-05-11 à 09 04 45
Nice !

knopers8:
I confirmed it works at P2, will resolve the conflicts soon.

kostorr (approved):

PR #329, opened 2022-05-10
PHS: Add  phos-raw-clusters-epn workflow to generate-all-workflows scripts

sevdokim:
Hello! 
It appeared that scripts/phos-raw-clusters-epn.sh was missing in scripts/generate-all-workflows.sh and was not regenerated in the flpsuite v0.57.0. This commit fixes the issue.


sevdokim:
hi @knopers8 can you please review this PR?it will be perfect if it can appear in flpsuite 0.58

PR #332, opened 2022-05-11
MFT: Adding the noise QC to the ECS options.

tomas-herman:


knopers8 (dismissed):

tomas-herman:
Hi @knopers8, to create this mft-noise-qcmn workflow I have basically copied the mft-full-qcmn workflow and renamed it to noise. 
I created the json file and the shell script in a previous PR: https://github.com/AliceO2Group/ControlWorkflows/pull/273
- https://github.com/AliceO2Group/ControlWorkflows/blob/master/scripts/etc/mft-noise-qcmn.json
- https://github.com/AliceO2Group/ControlWorkflows/blob/master/scripts/mft-noise-qcmn.sh

But I forgot to modify the `readout-dataflow.yaml` file so that is why I created this PR.

Is there some other step needed in order to add a workflow? Do I need to create some more files?

PR #340, opened 2022-05-19
regenerate DPL workflows for O2 nightly-20220519.1

vascobarroso:


knopers8 (commented):
@ktf There is just one workflow template which changed with the latest O2 tag, could you please confirm it is expected? This looks like some reordering.

Here is the command: https://github.com/AliceO2Group/ControlWorkflows/blob/492fc71f8de0a55f2b5911cc1d110b92a50238a7/scripts/cpv-pedestal-calib-qc.sh#L19

ktf:
That looks reasonable, I guess they added the DSFT bit.

PR #349, opened 2022-05-27
ITS: change publishing rate

JianLIUhep:
This can be merged into the current default revision at P2. Thank you.


JianLIUhep:
Hi @knopers8, could we modify the object publishing frequency by changing the parameter in the json instead of regenerating the workflow? It seems that simply changing the parameter did not work. Please let me know if I overlooked anything. Thank you.

Barthelemy:
@JianLIUhep  Unfortunately we cannot. The timer's period is set in the task definition. This will be solved by having Just In Time workflow generation.

Barthelemy:
@JianLIUhep Please, also make sure to apply the same changes in Consul

vascobarroso (approved):

PR #352, opened 2022-06-02
add tof match QC

noferini:
@Barthelemy Could you enable my repo in AliECS so that we can run a test?

### DIFF
--- a/scripts/etc/tof-full-epn-qcmn.json
+++ b/scripts/etc/tof-full-epn-qcmn.json
@@ -3,26 +3,89 @@
     "config": {
       "database": {
         "implementation": "CCDB",
-        "host": "ccdb-test.cern.ch:8080",
+        "host": "ali-qcdb.cern.ch:8083",
         "username": "not_applicable",
         "password": "not_applicable",
         "name": "not_applicable"
       },
       "Activity": {
-        "number": "42",
-        "type": "2"
+        "number": "0"
       },
       "monitoring": {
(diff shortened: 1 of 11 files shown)


noferini:
Thanks, I will report here the result of the test as soon as we did it

noferini:
Hi Bart,
qc-merger started properly and new QC merging task were added (I cannot fill them since epn workflow is still missing).
We can merge this PR.
Note that this change affect only the "emergency" configuration with compressors on epn.
At this stage we preferred not to touch the official one (we will do after the pilot beam)

Barthelemy:
ok very good

Barthelemy:
I have to let @vascobarroso or @knopers8 do it .

vascobarroso (approved):

vascobarroso:
@noferini is this for FLP Suite v0.60.0 (to be deployed at P2 next Tuesday) or for now ?

noferini:
@vascobarroso 
I don't know whatt are the plan for this week.
If @shahor02 and @davidrohr will manage to run tracking and matching on EPN, this PR will provided an alternative configuration to get the QC of the matched TOF-track in sync phase.
If possible/needed I would merge it to be ready in such case.
As mentioned, this PR doesn't touch the standard configuration (to avoid surprise).

vascobarroso:
@noferini I have pushed this to the current default revision being used at P2 (after the agreement of RC).

PR #358, opened 2022-06-02
Execute ccdb.RunStart earlier in the SOR sequence

vascobarroso:


vascobarroso:
@davidrohr @shahor02 this is the change to bring the creation of the GRP-ECS object earlier in the SOR sequence. Will become available at P2 next Tuesday.

knopers8:
@vascobarroso Was it tested? We should be sure that run number exists at `before_START_ACTIVITY`

shahor02:
Hi @knopers8 

Could you please add an execution of 
```
curl[ "http://o2-ccdb.internal/latest/%5Cw%7B3%7D/.*/`date +%s`000?prepare=true"](http://o2-ccdb.internal/latest/%5Cw%7B3%7D/.*/%60date+%25s%60000?prepare=true)
```
right after the call of `o2-ecs-grp-create`?

PR #360, opened 2022-06-03
Regenerate workflows for FLP Suite v0.60.0 + few fixes

Barthelemy:
use the name of the task instead of the field "taskName".


Barthelemy:
@vascobarroso @mfasDa

vascobarroso (approved):

PR #366, opened 2022-06-09
[OCTRL-665] Add DPL_OUTPUT_PROXY_ORDERED to each dpl-output-proxy

knopers8:


knopers8:
it will have an actual effect on templates after the next template regeneration

vascobarroso (approved):

vascobarroso:
@knopers8 thanks a lot! I will use this to regenerate the workflows for FLP Suite v0.61.0.

PR #370, opened 2022-06-10
CPV: fix for cpv compressor workflow

sevdokim:
Hi! This is fix for cpv compressor workflow.

### DIFF
--- a/scripts/etc/cpv-qc-compressor.json
+++ b/scripts/etc/cpv-qc-compressor.json
@@ -33,7 +33,7 @@
         "": "The other type of dataSource is \"direct\", see basic-no-sampling.json.",
         "dataSource": {
           "type": "direct",
-          "query": "peds:CLP/CPV_Pedestals;feethrs:CLP/CPV_FEEThrs;deadchs:CLP/CPV_DeadChnls;highthrs:CLP/CPV_HighThrs;pedeffs:CLP/CPV_PedEffs"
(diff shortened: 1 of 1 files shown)


tasks/cpv-compressor-dpl-output-proxy.yaml
```diff
@@ -25,7 +25,6 @@ command:
   env:
     - O2_DETECTOR={{ detector }}
     - O2_PARTITION={{ environment_id }}
-    - DPL_OUTPUT_PROXY_ORDERED=1
```
knopers8, line 28: this should stay, it is required by DPL to correctly order DPL output proxy

vascobarroso: @sevdokim which O2 tag do you have on the machine that you use to regenerate ? I think this new env var requires `nightly-20220605` or newer.

sevdokim: hi @vascobarroso 
I use flp162 to regenerate. So I think the version is that which corresponds to flp suite 0.60. Ok, problem is understood. Now I just add it manually but after next update at p2 we will see if it added automatically.

sevdokim: @knopers8 @vascobarroso can you please look now?

knopers8 (approved):

PR #373, opened 2022-06-13
TPC- Integrated digital current workflow

rmunzer:


scripts/tpc-idc.sh
```diff
@@ -0,0 +1,239 @@
...
+   --o2-control $WF_NAME
+
+
+#   --default-port 47734 \
+#   --tpc-idc-merger-proxy '--channel-config "name=tpc-idc-merger-proxy,method=connect,type=push,transport=zeromq" ' \
+
+
+# add the templated CRU config file path
+ESCAPED_CRU_FINAL_CONFIG_PATH=$(printf '%s\n' "$CRU_FINAL_CONFIG_PATH" | sed -e 's/[\/&]/\\&/g')
+sed -i /defaults:/\ a\\\ \\\ "${CRU_CONFIG_PARAM}":\ "${ESCAPED_CRU_FINAL_CONFIG_PATH}" workflows/${WF_NAME}.yaml
```
knopers8, line 81: FYI, `helpers.sh` contains a helper to do this, we just did not widely adopt it due to lack of time :)

workflows/tpc-idc-full.yaml
```diff
@@ -0,0 +1,9 @@
+name: tpc-calib-full
+roles:
+  - name: tpc-idc
+    enabled: "{{ it != 'alio2-cr1-flp145' }}"
+    include: tpc-idc
+  - name: tpc-current
+    enabled: "{{ it == 'alio2-cr1-flp145' }}"
+    include: minimal-dpl
```
knopers8, line 8: the names I proposed here were from the top of my head, if they are incorrect, feel free to change them.

scripts/tpc-idc.sh
```diff
@@ -0,0 +1,239 @@
...
+sed -i /defaults:/\ a\\\ \\\ "merger_port":\ "${PORT}" workflows/${WF_NAME}.yaml
+
+
+ORIGINAL_STRING="tpc-idc-merger-proxy-{{ it }}"
+REPLACE_STRING="tpc-idc-merger-proxy"
+
+sed -i "s/""${ORIGINAL_STRING}""/""${REPLACE_STRING}""/g" workflows/${WF_NAME}.yaml tasks/${WF_NAME}-*
+
+
+add_qc_remote_machine_attribute workflows/${WF_NAME}.yaml alio2-cr1-qts01
```
knopers8, line 227: this line should not be needed anymore after this: https://github.com/AliceO2Group/ControlWorkflows/pull/328

knopers8 (commented):
Please add the new script to `generate-all-dpl-workflows.sh`

workflows/tpc-idc-full.yaml
rmunzer: I changed the tpc-current since the official news will be likel tpc-sac (sampled analog currents)

scripts/tpc-idc.sh
rmunzer: I will have a look an include this function in the script.

scripts/tpc-idc.sh
rmunzer: Do i need to set the node name somewhere in consul or could i just remove the line?

rmunzer:
Hi @knopers8,

i add the file and did some correction.

scripts/tpc-idc.sh
```diff
@@ -0,0 +1,239 @@
...
+
+export GLOBAL_SHMSIZE=$(( 16 << 30 )) #  GB for the global SHMEM
+PROXY_INSPEC="x:TPC/RAWDATA;dd:FLP/DISTSUBTIMEFRAME/0"
+#PROXY_INSPEC="x:TPC/RAWDATA"
+
+OUTSPEC_IDC="idc2:TPC/IDCGROUP"
+OUTSPEC="xout:TPC/RAWDATA;ddout:FLP/DISTSUBTIMEFRAME/0"
+#OUTSPEC="xout:TPC/RAWDATA"
+# TODO: Adjust path to pedestal file
+pedestalFile="/home/tpc/IDCs/FLP/Pedestals.root"
```
Barthelemy, line 24: I don't think that this variable is used, or is it used in your code ? is it going to break when we run on a machine without tpc user ?

Barthelemy (approved):
Just one comment but I guess that it is fine like that. I let you confirm and then I merge.

scripts/tpc-idc.sh
rmunzer: Indeed this variable is not needed anymore. I will remove it from the script.

knopers8 (approved):

PR #380, opened 2022-06-22
[EMCAL-711] Add remote workflow for RawDataErrors

mfasDa:
Keep remote workflow for raw data errors separate for
the commissioning phase (as emc-qcmn-epnall-rawerrors-
remote). Once the workflow is tested and consul is adapted
it will be merged into emc-qcmn-epnall-remote and this
workflow will be removed again.

### DIFF
--- a/scripts/emc-qcmn-epnall-rawerrors.sh
+++ b/scripts/emc-qcmn-epnall-rawerrors.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+# set -x;
+set -e;
+set -u;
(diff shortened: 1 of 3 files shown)


Barthelemy (approved):

mfasDa:
For the record: Building the workflows has been tested locally and works fine. Input type should be "direct" as it should subscribe to all raw data error objects as we want to be notified whenever an error occurred, not just the errors in a tiny subsample of the data. Usually these are rare cases and the error information is tiny.

PR #394, opened 2022-06-28
TOF cleanup

noferini:
1) removing all checkers on QC digit inputs till the crashes are understood
2) adding more trending plots
3) adding TOF matching QC in merger

1,2) validated with a synthetic run
3) run in a synthetic global run but no plots receieved -> task validated, json to be tuned once workflow available in official repo


Barthelemy (approved):
As you are renaming `MatchingTOF` , could we envisage to delete the objects in this path in the QCDB ? or do you want to keep them ?

PR #401, opened 2022-07-01
[EMCAL-711] Enable raw error task in all EPN-based workflows

mfasDa:
- Enable raw error task in all EPN-based workflows
  as remote workflow for the QC node
- Remove commissioning workflow


mfasDa:
@Barthelemy @knopers8 Please review!

Barthelemy (approved):

PR #418, opened 2022-07-17
MUON: add new MCH+MID matched track task (and update standalone MCH track task)

aphecetche:
@aferrero2707 this PR is a slight change compared to yours : instead of changing the existing QC node workflow, I'm adding a new one, in order to have an easy fallback solution in case something is wrong with the new one...

The new QC node workflow (mch-qcmn-epn-full-track-matching-remote) uses 3 tasks : 

- MCH/MO/MCHTracks : the current one, for reference and backup if needed, that plots quantities from MCH standalone tracks
- GLO/MO/MCHStdTracks : the equivalent of the previous one, but with a new code (code sharing with next task) + a (rather important) fix for the cluster map
- GLO/MO/MUONTracks : same set of plots as above, but for MCH+MID matched tracks

The corresponding Consul configuration is called ... `mch-qcmn-epn-full-track-matching`


scripts/etc/mch-qcmn-epn-full-track-matching.json
```diff
@@ -0,0 +1,229 @@
...
+        "maxNumberCycles": "-1",
+        "dataSource": {
+          "type": "direct",
+          "query": "digits:MCH/DIGITS;orbits:MCH/ORBITS"
+        },
+        "location": "local",
+        "mergingMode": "entire",
+        "localControl": "odc",
+        "localMachines": [
+          "epn"
```
knopers8, line 41: This is OK, but please make sure that `o2-qc` is exeucted with `--host epn` on the EPNs.

scripts/etc/mch-qcmn-epn-full.json
```diff
@@ -35,7 +35,7 @@
           "query" : "digits:MCH/DIGITS;orbits:MCH/ORBITS"
         },
         "location": "local",
-        "mergingMode": "delta",
+        "mergingMode": "entire",
```
knopers8, line 38: This is probably not what you want. If you use `entire`, you need sources which can be uniquely identified, thus have the list of all EPNs in `localMachines` and put the currently used as the argument for `--host`. Is there a reason to avoid deltas in your case.
This being said, the value in the config file in consul will have the effect here.

Documented here:
https://github.com/AliceO2Group/QualityControl/blob/master/doc/Advanced.md#multi-node-setups

knopers8 (commented):
I think it should work as it is, but please check my comments.

scripts/etc/mch-qcmn-epn-full-track-matching.json
aphecetche: actually in the Consul configuration (`o2/components/qc/ANY/any/mch-qcmn-epn-full-track-matching`) we use : 

```
"localMachines" : [ "localhost", "epn"]
```

is that OK ?

scripts/etc/mch-qcmn-epn-full.json
aphecetche: hum, so there is some misunderstanding on our (my at least ;-)) side here I guess.

We're using `entire` in the consul configuration that is being used currently, as we noticed previously an issue (with our tracks task) with `delta` : the histograms were a cumulative sum. We understand this is because we don't have implemented the `reset`  method in our task(s) (and this is because reset at each cycle is not convenient when testing locally). 

From your comment here I now seem to understand there's no way out and we somehow *have to* use `delta` *and* implement the `reset` method to get what we want at Pt2. Or am I missing something here ? @aferrero2707

aferrero2707: I was looking again at the documentation as well, and that's also now my understanding. Sorry, I completely missed the fact that `entire` requires the full list of EPNs to be explicitly given, which I guess is not feasible.
This also explains why the QC plots are only sparsely filled at P2 since we changed the merging mode to `entire`.

So I propose the following strategy:
- put back `delta` in consul, such that at least the QC plots are filled with data from all EPNs
- implement the histogram reset at each cycle by default (this will require a SW update)
- add a task parameter to disable the reset when running the QC locally on a single machine

If we agree on this I will start immediately to implement the reset at each cycle. @knopers8 @aphecetche what do you think?

aphecetche: I agree to the first 2 points.

For the last point I wonder whether it's really needed (as this will require to add a task parameter to all our tasks). 
At least in my typical usage of the QC on a single machine, what I'm really interested in is the have the plots integrated over the full stat of the running, i.e. don't really care about cycles there, mainly want a single (giant) cycle.

aferrero2707: I just checked, and the `Reset()` method is actually already instrumented for all tasks except the one for tracks:
https://github.com/AliceO2Group/QualityControl/blob/master/Modules/MUON/MCH/src/PhysicsTaskDigits.cxx#L491
https://github.com/AliceO2Group/QualityControl/blob/master/Modules/MUON/MCH/src/PhysicsTaskPreclusters.cxx#L580
https://github.com/AliceO2Group/QualityControl/blob/master/Modules/MUON/MCH/src/DecodingErrorsTask.cxx#L321

aphecetche: ok, that's on me then. I will change the track task.

aphecetche: see https://github.com/AliceO2Group/QualityControl/pull/1355

scripts/etc/mch-qcmn-epn-full-track-matching.json
knopers8: Yes, but please put the same values here.

AFAIK, EPNs still use `--host localhost` for the merged workflows, so you probably need `localhost` as well :/

scripts/etc/mch-qcmn-epn-full.json
knopers8: I agree with the above.

knopers8: On a single machine (no mergers as well), the Reset method won't be invoked at each cycle. This is only done for merging deltas and enabling moving windows.

scripts/etc/mch-qcmn-epn-full-track-matching.json
aphecetche: @knopers8 @aferrero2707  sorry, missed that comment. Will add "localhost" to the json momentarily...

aphecetche: done, and also reverted to delta here also.

knopers8 (approved):

PR #423, opened 2022-07-19
add new PDP option to force run as RAW

vascobarroso:


vascobarroso:
@shahor02 this PR adds a new PDP expert button in the AliECS GUI to force a run to be classified as RAW. Still needs a change in the AlIECS core to push it to the EPN tasks as FairMQ option, should come with the next FLP Suite.

shahor02:
@vascobarroso I assume this option should override our data registration policy https://alice.its.cern.ch/jira/browse/GO-87?focusedCommentId=288150&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-288150 and require data registration as `raw`. 
Does is set some FMQ property?

vascobarroso:
@shahor02 exactly. The idea is that this new variable is set as a FMQ property by AliECS and then on your side you can react accordingly.

shahor02:
Is the name of property `pdp_force_run_as_raw`?

shahor02:
`force_run_as_raw` is fine for me, once confirmed, will add it to CTF storage policy

teo:
I'm implementing the variable `force_run_as_raw`. Please note that the target FairMQ device will receive the string `"true"` or `"false"`, and not the actual boolean values.

teo (approved):

PR #426, opened 2022-07-19
[EMCAL-792] Enable checkers for RawErrorTask for all histograms

mfasDa:
Checking presence of error code and leaving detailed
messages about found error codes in the InfoLogger.
Checks are performed on all raw error histograms.
Checkers are enabled for remote workflows of all
workflows including processing on the EPN.


mfasDa:
@knopers8 Please review!

knopers8 (approved):

PR #427, opened 2022-07-19
Update mid-full-qcmn.json

ValerieRamillien:
remove TraskQCCheck part

PR #419, opened 2022-07-18
remove TraskQCCheck part

ValerieRamillien:
change mid-flp-raw_epn_full-qcmn.json to mid-flp-raw_epn-qcmn.json to remove TraskQCCheck part

PR #439, opened 2022-08-03
Keep full MID QC configuration but disable track check

dstocco:
This PR is equivalent to https://github.com/AliceO2Group/ControlWorkflows/pull/419 and https://github.com/AliceO2Group/ControlWorkflows/pull/427 but, instead of pointing to a new json file on consul, it simply expects that the CheckMIDTracks is set to inactive.
This is expected to minimize the code modification when we will recover the check in the future.
The commits are meant to be atomic: please do not squash

### DIFF
--- a/scripts/etc/mid-full-qcmn.json
+++ b/scripts/etc/mid-full-qcmn.json
@@ -161,6 +161,20 @@
             "MOs": []
           }
         ]
+      },
+      "QcCheckMIDTracks": {
+        "active": "false",
+        "className": "o2::quality_control_modules::mid::TracksQcCheck",
+        "moduleName": "QcMID",
+        "detectorName": "MID",
+        "policy": "OnAny",
+        "dataSource": [
+          {
+            "type": "Task",
+            "name": "QcTaskMIDTracks",
+            "MOs": []
+          }
+        ]
       }
     }
   },
--- a/scripts/mid-full-qcmn.sh
+++ b/scripts/mid-full-qcmn.sh
@@ -8,7 +8,7 @@ source helpers.sh
 
 WF_NAME=mid-full-qcmn-local
 QC_GEN_CONFIG_PATH='json://'$(pwd)'/etc/mid-full-qcmn.json'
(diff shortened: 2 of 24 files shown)


scripts/etc/mid-full-qcmn.json
```diff
@@ -161,6 +161,20 @@
             "MOs": []
           }
         ]
+      },
+      "QcCheckMIDTracks": {
```
knopers8, line 165: I would expect that adding a new check affects the check runner name. To confirm, have you regenerated the workflow template after editing the config file?

It is also fine we leave it as it is, because we run all the regeneration scripts at each release, but `master` might not be usable for this workflow before this.

dstocco: Hi @knopers8 , what I did is simply to put the Check back in the configuration but with the "active" flag set to "false". If I understand it correctly, this is equivalent of removing it from the configuration.
And indeed, when I regenerated the files (see second commit of this PR), there is no change in the name since the flag excludes the check.
The re-generation only modifies: --fairmq-ipc-prefix. I guess because I'm generating it under Mac OSX and not under linux. By the way, if you know a way to force setingt the --fairmq-ipc-prefix to "@" as it is done under linux, it would be great.
Thanks!
Cheers,
Diego

knopers8: ah sorry, i did not notice it was set to `false`.

regarding the `--fairmq-ipc-prefix`, perhaps you can just use it as an argument when generating the template?

knopers8 (approved):

scripts/etc/mid-full-qcmn.json
dstocco: I think I found it. This prefix is set in O2. For linux, it is set to "@", for other systems, it is set to TMPDIR if it exists. So it is enough to do:
`TMPDIR="@" ./mid-calib-qcmn.sh`
I write it here for future reference.

PR #443, opened 2022-08-04
[EMCAL-767] Enable QC montitoring of EMCAL online calibration

mfasDa:


mfasDa:
@knopers8 Please review!

scripts/etc/emc-qcmn-epn.json
```diff
@@ -125,7 +125,27 @@
...
+				],
+				"initTrigger": [
+					"SOR"
+				],
+				"updateTrigger": [
+					"newobject:ccdb:EMC/Calib/TimeCalibParams"
+				],
+				"stopTrigger": [
+					"EOR"
+				]
```
knopers8, line 146: SOR and EOR triggers are not implemented yet. Since this task will be ran along the rest of the environment, you can safely use `"userorcontrol"` instead. The same applies for the other postprocessing tasks.

The value in consul will have the priority here, but I guess you would like to keep them in line.

knopers8 (approved):

PR #444, opened 2022-08-05
[EMCAL-767] Add stand-alone QC server workflow for EMCAL Calibration monitoring

mfasDa:


mfasDa:
@knopers8 Please review! This is the option that we run the calib QC workflow in a similar way as done by the TPC. We would prefer having this as first step in the commissioning phase for testing the new postprocessing workflow stand-alone before we merge it with our standard QC workflow.

workflows/qc-server-workflow.yaml
```diff
@@ -16,6 +16,7 @@ defaults:
       - tpc-qc-post-trending-remote
       - tpc-qc-post-calib-remote
       - tpc-qc-post-processing-remote
+      - emc-qc-post-calib-remote
```
knopers8, line 19: please use alphabetical order (but keep "none" first)

knopers8 (changes requested):
please add the new script to `generate-all-dpl-workflows.sh`

knopers8 (approved):

PR #445, opened 2022-08-05
Tpc add trending and checkers

rmunzer:


scripts/generate-all-dpl-workflows.sh
```diff
@@ -114,8 +114,8 @@ check_pwd
 ./tpc-full-calib-trend-qcmn.sh
 ./tpc-full-qcmn.sh
 ./tpc-full-nodummy-qcmn.sh
-./tpc-full-nodummy-noraw-qcmn.sh
-./tpc-full-nodummy-nopid-qcmn.sh
```
knopers8, line 118: if you are removing these two from here, please consider removing the scripts themselves.

scripts/generate-all-dpl-workflows.sh
```diff
@@ -114,8 +114,8 @@ check_pwd
...
+./tpc-full-nodummy-qcmn-pp.sh
+./tpc-full-nodummy-qcmn-pp-v2.sh
 ./tpc-full-nodummy-qcmn-pp.sh
```
knopers8, line 119: `./tpc-full-nodummy-qcmn-pp.sh` is there twice

scripts/generate-all-dpl-workflows.sh
rmunzer: Thanks. Done

scripts/generate-all-dpl-workflows.sh
rmunzer: Thanks. Done

knopers8 (approved):

PR #446, opened 2022-08-08
MFT: Adding check on increasing of entries to all QC tasks for 1 plot and adding new trending plots

tomas-herman:
- Check on increasing entries added to 1 plot from each task
- Added trending plots for Cluster and Digit for empty chips
- Added trending plot slices for H0 and H1 of MFT for digit occupancy
- Added trending plot slices for cluster pattern ID mean and std dev in 3 ranges of ID


knopers8 (approved):
please remember to update the corresponding files in consul upon the next flp suite deployment

PR #462, opened 2022-08-17
Resolve config path + WFT API update

teo:
Merge after https://github.com/AliceO2Group/Control/pull/415
Do not move this to a `flp-suite-*` branch before AliECS v0.50.0.

### DIFF
--- a/workflows/readout-dataflow.yaml
+++ b/workflows/readout-dataflow.yaml
@@ -1330,7 +1330,7 @@ roles:
         roles:
           - name: "stfb"
             vars:
-              dd_discovery_stfb_id: stfb-{{ ctp_readout_host }}-{{ util.NewID() }} # must be defined for all stfb roles
+              dd_discovery_stfb_id: stfb-{{ ctp_readout_host }}-{{ uid.New() }} # must be defined for all stfb roles
             connect:
               - name: readout
                 type: pull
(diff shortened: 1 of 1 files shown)


workflows/readout-dataflow.yaml
```diff
@@ -1385,12 +1385,12 @@ roles:
       var: it
     vars:
       detector: "{{DetectorForHost( it )}}"
-      readout_cfg_uri_standalone: "consul-ini://{{ consul_endpoint }}/o2/components/readout/ANY/any/readout-standalone-{{ it }}"
-      readout_cfg_uri_stfb: "consul-ini://{{ consul_endpoint }}/o2/components/readout/ANY/any/readout-stfb-{{ it }}"
+      readout_cfg_uri_standalone: "consul-ini://{{ consul_endpoint }}/o2/components/{{config.ResolvePath('readout/' + run_type + '/any/readout-standalone-' + it)}}"
+      readout_cfg_uri_stfb: "consul-ini://{{ consul_endpoint }}/o2/components/{{config.Resolve('readout', run_type, 'any', 'readout-stfb-' + it)}}"
```
teo, line 1389: @pippohub as you can see this is now happening, though not in MVS as I previously hinted since I realized code changes were necessary after all.

This will be ready to test "plug&play" after AliECS v0.50.0 is released and deployed to MVS (so next Suite). Then it is simply a matter of putting config files under `PHYSICS`, `TECHNICAL`, `ANY`, etc. and watch all config queries fall back to `ANY` automatically if a config file for a specific run type doesn't exist in Consul.

PR #470, opened 2022-08-25
[WIP] TRD and MCH remote workflows get 20GB shmem segment

knopers8:
this is to see if expanding the shared memory segment helps
with mitigating the Mergers becoming stuck not being able to create a message.


knopers8:
will merge once i see the problem we are tackling also with the current flp suite and only after RC approval

knopers8:
superseded by https://github.com/AliceO2Group/ControlWorkflows/pull/474

PR #474, opened 2022-08-30
[QC-845] Allow to set shmem segment size of remote QC per detector

knopers8:
This will allow to allocate more memory for workflows which require so.


knopers8:
tested on staging

vascobarroso (approved):

PR #480, opened 2022-08-31
[OCTRL-703] Allow to use JIT in qc-server-workflow

knopers8:
tested on staging, ready to go

### DIFF
--- a/workflows/qc-server-workflow.yaml
+++ b/workflows/qc-server-workflow.yaml
@@ -30,6 +30,8 @@ defaults:
       - alio2-cr1-qc01
       - alio2-cr1-qc02
       - alio2-cr1-qc03
+      - alio2-cr1-qc04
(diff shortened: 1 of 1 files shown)


workflows/qc-server-workflow.yaml
```diff
@@ -30,6 +30,8 @@ defaults:
       - alio2-cr1-qc01
       - alio2-cr1-qc02
       - alio2-cr1-qc03
+      - alio2-cr1-qc04
```
vascobarroso, line 33: @knopers8, `alio2-cr1-qc04` belongs to STAGING. Should it be on this list ?

knopers8: Yes, so we can use the same workflow on STAGING. I am not sure if we can display a different list depending on the setup.

vascobarroso (approved):

PR #482, opened 2022-09-05
[OCTRL-685] Trigger CCDB RunStop whenever we go to error

knopers8:
This will allow to publish EOR GRP also when we try to start a run, but we fail before reaching RUNNING.


knopers8:
needs https://github.com/AliceO2Group/Control/pull/420

teo (approved):

PR #484, opened 2022-09-06
[EMCAL-718] Enable monitoring of the bad channel map

mfasDa:


mfasDa:
@knopers8 Could you please have a look if this is OK? In particular I am wondering if I can add two triggers (one for each calibration object). Is there a way to check in the code which trigger was called?

scripts/etc/emc-qc-post-calib.json
```diff
@@ -29,13 +29,15 @@
...
+            "TimeCalibParams",
+            "BadChannelMap"
           ],
           "initTrigger": [
             "userorcontrol"
           ],
           "updateTrigger": [
-            "newobject:ccdb:EMC/Calib/TimeCalibParams"
+            "newobject:ccdb:EMC/Calib/TimeCalibParams",
+            "newobject:ccdb:EMC/Calib/BadChannelMap"
```
knopers8, line 40: This will basically make `update()` get triggered whenever one of the two objects is updated. There is no way to know which of the two fired, since they are the same type. Perhaps we could have the config string in the `Trigger` struct, if that helps.

knopers8: Let me know if I should merge it then. It is OK for me like this if you understand the implications

knopers8 (approved):
got the green light from @mfasDa on mm

PR #491, opened 2022-09-07
[EMCAL-719, EMCAL-767] Simple formatting fixes of json configs

mfasDa:
Fix formatting of two configs which got out-of-format using the VSCode json formatter


mfasDa:
Simple format fix, no change in code

knopers8 (approved):

PR #497, opened 2022-09-12
Forbid JIT for TPC FLPs until we can run TPC IDC workflow

knopers8:
WIP until I confirm it works on staging.


knopers8:
good to go

vascobarroso (approved):

PR #515, opened 2022-10-11
Add NOISE run type to drop down

vascobarroso:


shahor02:
@vascobarroso, is this a legitimate run type which can go to the GRPECS? 
In the Control I see multiple files referring on runTypes, could you please point me to the latest primary source of the runTypes? 
In the O2 we are using a continuous enum as a runType, while e.g. in Control/apricot/protos/apricot.pb.go I see a departure from this scheme, we may need to add a lookup table.

vascobarroso:
@shahor02 yes, this is a new run type that was requested by MFT. 

I will let @teo answer about the authoritative source of truth concerning run types inside Control. 
But I think we need to call soon for a meeting with the different stakeholders (RC, AliECS, DCS, PDP) to consolidate the run types in ALICE, since more requests will come.

teo:
Hi @shahor02,
I consider https://github.com/AliceO2Group/Control/blob/master/common/runtype/runtype.go as the source of truth on run types, with https://github.com/AliceO2Group/Control/blob/master/apricot/protos/apricot.proto being the Protobuf equivalent of the same plus a special value `ANY` we need for task configuration payload resolution in AliECS.
These two are also synchronised with the DCS interface protofile https://github.com/AliceO2Group/Control/blob/master/core/integration/dcs/protos/dcs.proto#L225 which includes a subset of the global list.

I keep `CALIBRATION`, `COSMICS` and `SYNTHETIC` after a gap in numbering because they are not supported by the DCS, so this allows DCS to e.g. add `NOISE` as the next enum value without having to skip or reindex to account for these "special" run types. I welcome a discussion to consolidate these values.

shahor02:
Thanks @teo !
I'll add the necessary changes in O2. Could you please ping me when the run types list is getting updated?

shahor02:
Hi @teo 

After more careful checking: redefinition https://github.com/AliceO2Group/Control/commit/c2120a6600d31bbdcc4ce6b476e32660e42fa6b3#diff-7921bfb7aaf0d6a70fdd5d68644f71903ce4942dcd1c3275d3a98dd924e5e93fR50-R52 create a problem: the runType is stored in the in GRP object and we have already plenty of objects with e.g. `Cosmic` stored as runType 14. Why DCS cannot support contiguous enumeration, simply considering dummy those IDs it does not process?

teo:
I cannot answer this - perhaps DCS could support it, I just know that when `COSMICS` came up it was decided to send `PHYSICS` to DCS since `COSMICS` wasn't a value in their system.

What if I set `COSMICS` back to 14, then `NOISE` gets 15, and `CALIBRATION` and `SYNTHETIC` stay with their high indexes (200 and 202 respectively) until we figure out what to do with them. Would this work for you?

shahor02:
@teo not sure this will work: we almost have not `CALIBRATION` GRPs stored but have a lot of `SYNTHETIC`.
I think we should simply avoid redefining already used IDs and add new ones. Why DCS cannot have special treatment internally for the types it does not support? Who is working on this from DCS?

teo:
@shahor02 to be clear I would want for nothing more, than for the run type list to be one and the same, and for DCS to use this same enum list and handle inernally the RT values it doesn't support. The DCS team has been invited to join the RT consolidation effort, and they have a new developer working on the DCS-ECS interface. I hope this time we'll sort it out for good.

shahor02:
@teo ok, let's discuss this at the corresponding meeting. But is the version with redefined IDs is already deployed? If so, working with the GRP of `synthetic` runs (which we usually don't do) may leed to a crash.

Barthelemy (approved):

PR #518, opened 2022-10-18
Split GLO for VTX and ITSTPC_Matching

chiarazampolli:
@knopers8 : note that I left the previous configuration till we converge on the idea of splitting.

### DIFF
--- a/jit/glo-itstpc-mtch-qcmn-remote
+++ b/jit/glo-itstpc-mtch-qcmn-remote
@@ -1,0 +1,1 @@
+o2-qc --config consul-json://{{ consul_endpoint }}/o2/components/qc/ANY/any/glo-itstpc-mtch-qcmn-epn --remote -b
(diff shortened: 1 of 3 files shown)


vascobarroso (dismissed):

knopers8 (commented):
I think you need to add the two workflows also in the selection panel:
https://github.com/AliceO2Group/ControlWorkflows/blob/master/workflows/readout-dataflow.yaml#L831

chiarazampolli:
Hi @knopers8 , 
Like this? I removed the previous "value" because I want that only the two new ones are used. 
One question: I also left "none", but I don't know what it means.
Cheers,
Chiara

knopers8 (approved):

knopers8:
> Hi @knopers8 , Like this? I removed the previous "value" because I want that only the two new ones are used. One question: I also left "none", but I don't know what it means. Cheers, Chiara

Yes, like this. "none" means that no workflow is created for given detector/subsystem. If you have a look at the AliECS GUI and the FLP workflows panel, I hope it becomes quite obvious.

PR #519, opened 2022-10-18
Tpc idc 5

rmunzer:
Add idc and sac workflow for direct connection to calib node


Barthelemy (approved):
Given the complexity of the scripts, I have to trust you on this. Nevertheless I had a long look and I don't see obvious mistakes.

PR #523, opened 2022-10-26
Improvements to PDP QC settings

davidrohr:
To be deployed with additional changes as described in https://alice.its.cern.ch/jira/browse/OCTRL-736


vascobarroso (approved):

workflows/readout-dataflow.yaml
```diff
@@ -1094,19 +1085,19 @@ defaults:
     index: 600
     visibleif: $$epn_enabled === "true" && ($$pdp_config_option === "Repository hash" || $$pdp_config_option === "Repository path") && $$pdp_panel_mode === "Expert"
   pdp_detector_list_qc: !public
```
teo, line 1087: As discussed in https://alice.its.cern.ch/jira/browse/OCTRL-736 this variable and the next one don't exist any more, because new variables that act as exclude lists were added instead.

PR #533, opened 2022-11-25
[OCTRL-704] Enable JIT for TPC IDC workflows

knopers8:


knopers8:
tested at P2, works fine

vascobarroso (approved):

PR #545, opened 2023-02-17
restore old TOF compressor behaviour

noferini:


noferini:
Hi @knopers8 , @Barthelemy ,
we found that the last change in TOF compressor doesn't work properly.
By additin "--old" flag we can restore old behaviour (tested this morning -> working).
Can you port this change in the current suite?

PR #548, opened 2023-02-22
Split task stdout/stderr logging settings

teo:


teo:
@knopers8 this requires a change in the DPL output, or alternatively setting `log_task_output` to some value in Consul so that generated subworkflows still work.

tasks/fairmq-shmmonitor.yaml
```diff
@@ -14,7 +14,8 @@ wants:
   cpu: 0.01
   memory: 3
 command:
-  log: "{{ log_task_output }}"
+  stdout: "{{ log_task_stdout }}"
+  stderr: "{{ log_task_stderr }}"
```
knopers8, line 18: since you are getting rid of `log_task_output`, I would also remove the default value a few lines above and add defaults for ` log_task_stdout` and `log_task_stderr`. Who knows if we won't have new "mother" workflows, which may not have these two defined.

workflows/qc-server-workflow.yaml
```diff
@@ -74,14 +74,26 @@ defaults:
...
+    value: none
+    type: string
+    label: "Task stderr"
+    description: "Select to where the standard error of controlled tasks should be forwarded to (select 'all' for InfoLogger)"
+    widget: dropDownBox
+    panel: Logging
+    values:
+      - "none"
+      - "stdout"
+      - "all"
```
knopers8, line 87: I don't quite understand why are you still using `"none"`, `"stdout"` and `"all"` for both `log_task_stdout` and `log_task_stderr` instead of on/off buttons.

knopers8 (changes requested):

tasks/fairmq-shmmonitor.yaml
teo: Thanks, not sure how I missed that

workflows/qc-server-workflow.yaml
teo: Because it's not a boolean, it's a 3-state switch:
- none → the stream goes to `/dev/null`
- stdout → the stream goes to the executor's `stdout`
- all → the stream goes to all outputs, i.e. the executor's `stdout` as well as InfoLogger

Both the task's `stdout` as well as the task's `stderr` can take any of these values.

knopers8: I see. Why executor's `stdout` would be useful?

knopers8: OK, thanks for the explanation. I fear though it will be just as confusing to the users.

knopers8:
https://github.com/AliceO2Group/AliceO2/pull/10821 here is the O2 part

knopers8 (approved):

PR #552, opened 2023-03-07
add compressor workflow for 32 HBF

noferini:


noferini:
Removing rdh version 6 requirement (using default -> 7)
Adding one workflow for 32 HBF

PR #554, opened 2023-03-14
[OCTRL-770] Move CCDB run stop to after_STOP_ACTIVITY

knopers8:


vascobarroso:
@knopers8 what happens if e.g. we kill an ongoing run instead of stopping ? Will `ccdb.RunStop()` still be triggered ?

teo (approved):

PR #557, opened 2023-03-17
Parametrize StfSender SHM size

vascobarroso:
@teo is there any way of using a different value for the StfSender running in the CTP readout node ?

### DIFF
--- a/tasks/stfsender.yaml
+++ b/tasks/stfsender.yaml
@@ -13,6 +13,7 @@ defaults:
   log_task_stdout: none
   log_task_stderr: none
   stfs_dd_region_size: 4096
+  stfs_shm_segment_size: 33554432
   _module_cmdline: >-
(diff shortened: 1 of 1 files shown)


teo:
> @teo is there any way of using a different value for the StfSender running in the CTP readout node ?

Yes, we can just override the `stfs_shm_segment_size` value with a `vars` line within the `readout-ctp` role. Task defaults are more "defaulty" than WFT vars coming from above.

teo:
> @teo cool, could you then add that to this PR ?
> 
> To be clear, what we want is that we can set **in Consul** different values for standard FLPs and the CTP readout FLP. So I guess we would need separate var names ?

Values can now be read in from Consul as `stfs_shm_segment_size` and `ctp_stfs_shm_segment_size`, both default to `33554432`.

teo (approved):

PR #559, opened 2023-03-23
disable core dumps for StfSender

vascobarroso:


vascobarroso:
@teo I guess we could even go further and make it a runtime parameter for all tasks right ? 
That should allow us to decide (together with some system settings to be deployed by Ansible) if we want core dumps or not at runtime.

knopers8:
Well, then we probably should make it dependent on a variable. I can have a look when I have a moment.

PR #560, opened 2023-03-24
tof test modes

noferini:


noferini:
This is the last tuning of TOF workflow validated in the last days (1 cosmic run of 14h)

vascobarroso:
@noferini should this be added in next Monday's SW deployment ? Immediately to the current FLP Suite at P2 ?

noferini:
Hi @vascobarroso , this is required to join global run eith TOF. As soon as possible is the best for us

vascobarroso (approved):

vascobarroso:
@noferini just to be sure, this is compatible with the current O2 version at P2 right ?

PR #562, opened 2023-03-29
FLP FIT: workflows with rawdata sampling decreased number of pipelines

afurs:


afurs:
Number of decoder pipelines decreased 4 -> 1. Due to raw data sampling with 1% (QC can handle only this fraction of nominal data)

knopers8 (approved):

PR #568, opened 2023-04-17
set 15 compressor as default

noferini:


vascobarroso:
@noferini should this be merged and added to the current FLP Suite v0.94.0 at P2 ?

knopers8 (approved):

noferini:
It is not urgent, we can wait. Thanks

PR #567, opened 2023-04-16
setting DPL_PIPELINE to 32 for TOF flp workflow

noferini:
This is to set as default the option suggested by @ktf for tof flp worflow

PR #573, opened 2023-04-20
Revert "setting DPL_PIPELINE to 32 for TOF flp workflow (#567)"

vascobarroso:
This reverts commit 759484a1000520412c891b387b7aa9ec9f65bc2f.


vascobarroso:
@noferini this PR reverts the setting of DPL_PIPELINE to 32 in master so we don't get it back in the next releases. Could you check that it is ok (there were some conflicts because of the other change in https://github.com/AliceO2Group/ControlWorkflows/pull/568.

PR #574, opened 2023-04-21
[FOCAL-12] Pedestal DA workflow for Pads

mfasDa:
- Run pedestal workflow for pads
- Publish to CCDB on FLP
- Store ADC distribution in /tmp for development,
  testing and debugging


mfasDa:
Needs https://github.com/AliceO2Group/AliceO2/pull/11191

mfasDa:
@knopers8 Please review!

@shahor02 @chiarazampolli Do the CCDB populator settings look OK? For the moment we would think it is the best to use the CCDB on the FLP to store the pedestals. We don't interact with DCS. I took the template from the CPV pedestal run and replaced FOCAL-specific settings.

mfasDa:
Pinging also @ tommasoisi

teo (approved):

PR #575, opened 2023-04-26
[FOCAL-12] Include pedestal QC in pedestal DA

mfasDa:


mfasDa:
@teo @knopers8 Please review!

mfasDa:
Uses https://github.com/AliceO2Group/QualityControl/pull/1744

Barthelemy:
The config file is not in Consul at P2: foc-padpedestal-qc-*
Thus I cannot really say, but the modification simply adds QC to this workflow.

mfasDa:
> The config file is not in Consul at P2: foc-padpedestal-qc-*
> Thus I cannot really say, but the modification simply adds QC to this workflow.

The FOCAL FLP is configured as stand-alone FLP, thus the local consul instance is used. You find the configuration here: http://alio2-flp-focal.dyndns.cern.ch:8500/ui/alice-o2-cluster/kv/o2/components/qc/ANY/any/foc-padpedestal-qc-alio2-flp-focal/edit

Barthelemy (approved):

PR #577, opened 2023-05-05
Include minimal dpl workflows for FOC FLP

mfasDa:


mfasDa:
@teo I think this can be merged without problems. We also need the minimal workflows for testing and debugging

mfasDa:
Withdrawing PR

PR #584, opened 2023-05-26
[O2-3691] Use 'drain' in all the remote QC worklflows

knopers8:
@ktf as you suggested, but I doubt it will make any difference, because 'drain' is already the default.


ktf:
As far as I can tell, "keep" is the default. See:

Framework/Core/src/ExternalFairMQDeviceProxy.cxx:489

also, notice that's an option for the input proxy. Maybe that's the reason?

knopers8:
as said in the ticket, we actually do this to the input proxies in QC:
https://github.com/AliceO2Group/QualityControl/blob/db356f2fa6adc8c7370aabe25ccd13739499471f/Framework/src/InfrastructureGenerator.cxx#L72
so there is no need to set it anywhere else.

Closing the PR, as it would break the workflows by providing the --ready-state-policy argument on the workflow level, not data processor level.

PR #585, opened 2023-05-26
CPV: increase nTFs per slot in pedestal and noise calibration

sevdokim:
Hi! In this PR slot length for CPV noise and pedestal calibration is increased 4 times (from 100 to 400). This comes out from change of TF size from 128 HBFs to 32 HBFs.


sevdokim:
Hi @knopers8 
I have a question regarding possibility to have variable fetching from aliecs. Before jit it was possible with writing dpl command in way like 
```
...--tf-per-slot {{ tf-per-slot }}
```
and setting default value of `tf-per-slot` in the script. Now if I understand correctly it is also possible to use such variable but it needs to be setted everytime from aliecs. Is there a possibility to setup default value for some arbitrary variable with jit approach?

knopers8:
Hi, you can either:
- add the default value in consul in `o2/runtime/aliecs/defaults/`
- Put it in the template directly in the `Other variables`: https://github.com/AliceO2Group/ControlWorkflows/blob/7003bae0cbe862b966b51be4217b0f5b5e0ca489/workflows/readout-dataflow.yaml#L1347

However, please do not use minuses in `tf-per-slot`, since these might be treated as a subtraction statement. Also, please add a `cpv` prefix, so you don't overlap with a similar value from another detector. I.e. `cpv_tf_per_slot`

sevdokim:
I'll try a run and let you know. Don't merge yet please.

sevdokim:
Hi @knopers8 
thank you very much, it works. I used 2nd approach: added `cpv_ped_tf_per_slot` and `cpv_noise_tf_per_slot` into workflows/readout-dataflow.yaml to other variable section. You can merge.

knopers8 (approved):

PR #595, opened 2023-06-29
Revert "Disable raw parser counter check for IDCs (#592)"

vascobarroso:
This reverts commit b17b9d96b50c07374caaa095bcede66920d619bb.


vascobarroso:
@rmunzer PDP now has a proper SW fix for this so they requested that we revert this change since we anyway deploy the new SW tomorrow morning. Let me know if you object.

knopers8 (approved):

PR #597, opened 2023-07-07
Change TPC IDC & SAC proxy ports

knopers8:
The old ports are in the range of ephemeral ports on our machines, thus there is a risk that we might try to bind 47900 and 47901 when they are already used.


knopers8:
This should go together with https://github.com/AliceO2Group/AliceO2/pull/11626 to the same FLP suite release.

vascobarroso (approved):

PR #603, opened 2023-07-21
Lower CTP readout memory resource wants

teo:


knopers8:
@vascobarroso should we close it?

vascobarroso:
Yes, I think we can close it.

PR #607, opened 2023-08-10
Re-enable DCS PrepareForRun call

teo:
This reverts commit a22447f319b583ad2a55f36fde4da73c50bc2ccd.


vascobarroso:
@teo according to RC there is still one detector that does not support PrepareForRun in a given state, until that is fixed we cannot re-enable this. They hope to converge in the coming days, will let us know.

vascobarroso (approved):

PR #614, opened 2023-09-20
workflows for HI with lumi inputs: ZNC and VCH

lietava:
Two new ctp workflows with lumi inputs ZNC and VCH


knopers8 (changes requested):
please also add the workflows here, so you are able to select them in the GUI:
https://github.com/AliceO2Group/ControlWorkflows/blob/354a0ffb32094c861748f440cc261841266c258c/workflows/readout-dataflow.yaml#L133

aferrero2707:
@vascobarroso @knopers8 @lietava I would propose to add one more workflow with 6 pipelines, just in case we will be short in CPU cycles with high-rate HI:
* name: `jit/ctp-reco-6dec-HI`
* pipelines setting: `--pipeline ctp-raw-decoder:6`

Thanks!

knopers8 (approved):

lietava:
- names changed to reflect used inputs
- ctp-reco-6dec-HI-ZNC-TCE created
- new workflows added to readout-dataflow.yaml

PR #620, opened 2023-10-08
add tof paylaod limits

noferini:
This is to add a payload limit in TOF compressor raw-decoding.
A limit of 10M is added to all the default workflows (15cmp) with different settings.
An additional workflow was added 

tof-full-qcmn-cdump-bckp-local
as a backup of the previous tof-full-qcmn-cdump-15cmp-local version, in case we need to go back to previous behavior

We need this PR in master and at P2 as soon as possible (before of the today Pb-Pb runs)

We ran some tests this morning to see if the features we implemented work properly. All tests passed.


workflows/readout-dataflow.yaml
```diff
@@ -680,6 +680,7 @@ defaults:
       - tof-full-qcmn-paran-10cmp-local
       - tof-full-qcmn-paran-15cmp-local
       - tof-full-qcmn-cdump-15cmp-local
+      - tof-full-qcmn-cdump-bckp-local
```
knopers8, line 683: there is no such workflow in the master nor your PR, a leftover from some tests?

knopers8:
does it have the RC blessing to bring it to P2?

vascobarroso (approved):

workflows/readout-dataflow.yaml
noferini: You are right, I forgot to add it when I commit. I'm going to open a new PR since this is already merged

PR #622, opened 2023-10-13
add extra buffer protection in tof-compressor

noferini:
This is to add an extra protection to tof-compressor as suggested by @preghenella 

It adds an extra buffer before the encoding of compressed data, preventing to go out of memory in case the compressed size is larger than uncompressed (such a case is excluded for sure beyond the tolerance added here)
To be ported at P2 as soon as possible

@preghenella, can you also comment on this?


noferini:
In the meanwhile, I also ran some tests. We can merge it and port it to P2

preghenella:
looks ok to me

noferini:
Validated also at P2 in a cosmic run

noferini:
Since we collected one TF producing an error and this is not enough to protect I close this PR (we will add all protections in one O2 PR once ready)

PR #628, opened 2023-12-03
Add support for QC big screen

aferrero2707:
The QC Big Screen task creates a canvas with a summary of the detectors qualities, like this:
image

The QC Big Screen is enabled with a new checkbox in the QC Node Wokflows panel, associated to a dedicated JIT workflow, called `glo-qc-big-screen`:
image


aferrero2707:
@Barthelemy @knopers8 this PR introduces the workflow and GUI checkbox to enable the QC big screen with the detector qualities.
I have tested it in production with a synthetic run and it seems to b working properly.

Barthelemy (approved):

Barthelemy:
That's very good, thank you !

PR #630, opened 2023-12-07
Remove faulty ?lifetime specification

martenole:
@Barthelemy and @davidrohr we would need to at least revert this change to the dataspec which was not working. I remember now that in the QC json files I had to put also the **subspec** for the data if I wanted to add the `?lifetime` syntax. Can it be that this gave us problems? So that it should have been
```
--dataspec TF:MCH/RAWDATA_SAMPLED/0?lifetime=sporadic
```
instead of
```
--dataspec TF:MCH/RAWDATA_SAMPLED?lifetime=sporadic
```
Of course as mentioned in the meeting this should not be done for raw data but I'd still like to understand why we got these strange errors when trying to specify sporadic.


davidrohr:
ok, but this is processing rawdata_sampled, so it should indeed be sporadic, if I understand correctly.
I don't know how the syntax must be, never tried myself. But if you say it should be /0?lifetime=...., then I assume that is what we should put there?

martenole:
Ok, so I have tried now locally the last workflow which got changed: `jit/mft-noise-qcmn-local`. It indeed works to start the workflow when I put `--dataspec TF:MFT/RAWDATA_SMP/0?lifetime=sporadic`. But then also the `mft-digits-sampling` needs to be changed and the `/0` needs to be appended to the output, otherwise there is no match.
```
o2-dpl-raw-proxy -b --session default --dataspec 'x:MFT/RAWDATA;dd:FLP/DISTSUBTIMEFRAME/0' --inject-missing-data --readout-proxy '--channel-config "name=readout-proxy,type=pull,method=connect,address=ipc:///tmp/stf-builder-dpl-pipe-0,transport=shmem,rateLogging=10"' | o2-datasampling-standalone -b --config json://mft-digits-sampling.json | o2-itsmft-stf-decoder-workflow -b --runmft --digits --no-clusters --dataspec TF:MFT/RAWDATA_SMP/0?lifetime=sporadic --ignore-dist-stf
```
[mft-digits-sampling.json](https://github.com/AliceO2Group/ControlWorkflows/files/13601560/mft-digits-sampling.json)

davidrohr:
ok, thx for taking care!

martenole:
For reference the required changes in CONSUL:
```
diff --git a/o2/components/qc/ANY/any/mch-digits-datasampling b/o2/components/qc/ANY/any/mch-digits-datasampling
index 410522d1..359141e2 100644
--- a/o2/components/qc/ANY/any/mch-digits-datasampling
+++ b/o2/components/qc/ANY/any/mch-digits-datasampling
@@ -5,7 +5,7 @@
       "active": "true",
       "machines": [],
       "query": "raw:MCH/RAWDATA",
-      "outputs": "raw-sampled:MCH/RAWDATA_SAMPLED",
+      "outputs": "raw-sampled:MCH/RAWDATA_SAMPLED/0",
       "samplingConditions": [
         {
           "condition": "random",
diff --git a/o2/components/qc/ANY/any/mft-digits-sampling b/o2/components/qc/ANY/any/mft-digits-sampling
index 68b3fa55..0fb05d71 100644
--- a/o2/components/qc/ANY/any/mft-digits-sampling
+++ b/o2/components/qc/ANY/any/mft-digits-sampling
@@ -5,7 +5,7 @@
       "active": "true",
       "machines": [],
       "query": "raw:MFT/RAWDATA",
-      "outputs": "raw-sampled:MFT/RAWDATA_SMP",
+      "outputs": "raw-sampled:MFT/RAWDATA_SMP/0",
       "samplingConditions": [
         {
           "condition": "random",
```

Barthelemy:
Thank you very much !

PR #632, opened 2023-12-12
TPC IDC lifetime fix

martenole:
Please check @davidrohr and @matthias-kleiner 
I assume that also the `jit/tpc-idc-merger` requires a fix, but I was not sure how


martenole:
Hi @vascobarroso could this be merged please?

vascobarroso (approved):

vascobarroso:
@martenole done. Will be part of the next FLP Suite release, please let me know if it should be deployed sooner.

martenole:
Thanks @vascobarroso ! I think it would be good to have it in the current FLP suite. If there is time tomorrow or Friday for a global synthetic run we could try again without the `DPL_WORKAROUND_DO_NOT_CHECK_FOR_CORRECT_WORKFLOW_LIFETIMES` and see if that works then. If not, we might be able to provide another fix before the next update

vascobarroso:
@martenole this change is now at P2. I let you organize the test to check if it helps or not.

PR #635, opened 2024-01-26
Change pfr to non critical

rmunzer:


vascobarroso:
@rmunzer is this supposed to become official with the next FLP Suite release ? Or is it still under tests ?

teo (approved):

PR #638, opened 2024-03-14
MUON: add remote workflows for fwd matched tracks

aferrero2707:
This adds a new switch to enable/disable the QC node workflows for forward matched racks, similar to what is already done for ITS-TPC.

Three alternative workflows are selected depending on the list of detectors (MFT/MCH/MID) included in the data taking.

### DIFF
--- a/workflows/readout-dataflow.yaml
+++ b/workflows/readout-dataflow.yaml
@@ -891,6 +891,15 @@ defaults:
     panel: "QC_Nodes_Workflows"
     index: 21
     visibleif: $$qc_remote_enabled === "true"
+  glo_fwd_mtch_qc_enabled: !public
+    value: "false"
+    type: bool
+    label: "GLO Forward Matching QC"
+    description: "Enable/disable QC node workflow for forward Matching QC. Deployed only if MCH present!"
+    widget: checkBox
+    panel: "QC_Nodes_Workflows"
+    index: 22
+    visibleif: $$qc_remote_enabled === "true"
   glo_mean_vtx_calib_post_proc_qc_enabled: !public
     value: "false"
     type: bool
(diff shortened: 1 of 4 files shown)


workflows/readout-dataflow.yaml
```diff
@@ -1791,6 +1800,44 @@ roles:
               trigger: CONFIGURE
               timeout: 1s
               critical: false
+      - name: qc-remote-workflow-fwd # GLO is not a detector so we won't iterate on it with 'detectors' and we need special enable logic
+        enabled: "{{ 'MCH' in json.Unmarshal(detectors) }}"
+        vars:
+          qc_remote_workflow: "{{ util.PrefixedOverride( 'qc_remote_workflow', 'glo' ) }}"
```
aferrero2707, line 1806: @knopers8 it is ok to use the `glo` prefix for the workflow in two separate rules?

aferrero2707:
@vascobarroso @knopers8 @teo could you please have a look and check if it is all good? Thanks!

workflows/readout-dataflow.yaml
knopers8: Well, i don't think this variable is used for generating the actual workflow. It is probably a leftover of earlier approaches. 

The only potential issue I can see is that `qc_remote_workflow != 'none'` is used to enable shmem cleanup. While I see in a prod environment that it is not enabled, to be on the safe side, I would propose to remove the `fairmq-shmmonitor` from this role and from the role `qc-remote-workflow-glo`.

aferrero2707: @knopers8 thanks! I have removed the `fairmq-shmmonitor` from the two roles. I will organize a test in production with replay data to check that all is fine, and post here the outcome. The initial version was already checked in staging.

teo (approved):

knopers8 (approved):

aferrero2707:
@knopers8 @teo we just took one validation run in production ([548347](https://ali-bookkeeping.cern.ch/?page=run-detail&runNumber=548347)), seems to be working properly.

PR #641, opened 2024-04-03
[EMCAL-1118] Add QC node workflow for EMCAL pedestal calibration QC

mfasDa:


mfasDa:
@knopers8 @teo Please review and merge if OK!

knopers8:
Do you need it in prod?

knopers8 (approved):

mfasDa:
@knopers8 Yes, starting from the next FLP Suite release we should be ready to commission the pedestal run at P2. This PR relates to the dedicated QC task.

knopers8:
Ah, so by the next FLP suite release do you mean 1.27.0?

mfasDa:
Yes, because it still needs a commit in O2 which was merged this morning, and a commit in QualityControl which was merged after the release was tagged.

PR #642, opened 2024-04-03
[QC-1008] added fairmq header size settings to jit

justonedev1:


justonedev1:
Everything was tested on staging with all detectors on synth_pp_500khz workflow

knopers8 (approved):

PR #643, opened 2024-04-08
[OCTRL-880] remove obsolete cpv-qc-compressor

vascobarroso:


knopers8 (approved):

knopers8:
one could also remove `qc/ANY/any/cpv-qc-compressor` in consul assuming that other workflows do not refer to it

vascobarroso:
@knopers8 thanks for the approval and the suggestion. I have checked and could not find any other references to that config file. I would then suggest to remove it from consul once we deploy this in production.

PR #647, opened 2024-05-24
CTP: set max allowed input size to protect the decoder from ringing inputs

aferrero2707:
The `--max-input-size 700000` parameter has been added to the `o2-ctp-reco-workflow` in all the FLP workflows. The value is about 10% below the data size expected from ringing inputs.

Also added `--max-input-size-fatal true` to issue a fatal error in case large data input size is detected. The fatal error will trigger the stop of the ongoing run.


aferrero2707:
@lietava @vascobarroso can you have a look? Thanks!

lietava:
@aferrero2707 it looks ok for me.

knopers8 (approved):

vascobarroso:
@aferrero2707 should this be included in this week's deployment ? If so, I can cherry pick it into the release branch.

vascobarroso:
@lietava this was deployed yesterday so it's available in production

PR #649, opened 2024-06-04
Use alias for CTP FLP in production

vascobarroso:


vascobarroso:
@teo to simplify operations we are introducing an alias for the CTP FLP in production

`alio2-cr1-flp163` => `alio2-cr1-flpctp`

Will this work for AliECS ?

PR #651, opened 2024-06-10
parametrize StfSender dd-region-size for CTP FLP

vascobarroso:


vascobarroso:
@pippohub do we still need to be able to set different SHM buffer sizes for StfSender in regular FLPs and the CTP FLP ? 
I've checked in production and currently we have the same value everywhere. 

If we need to have different values we need this change since currently what we can set different is var `stfs_shm_segment_size` which as we have learned some time ago does not define the size of the SHM buffer (the good var is `stfs_dd_region_size`).

PR #653, opened 2024-07-04
CTP: added --local --host flp to the o2-qc commands on FLP

aferrero2707:
The `--local --host flp` options are needed in order to avoid running the post-processing and checker tasks on the CTP FLP.
The PR also adds the `?run_type={{ run_type }}` parameter to the Consul path for the remote workflow, allowing to enable/disable certain checks and post-processing tasks based on the run type (thanks to Apricot).


aferrero2707:
@vascobarroso @knopers8 we tested in staging the changes with @lietava and all seems to work as expected. Can you have a look and if ok include this PR in the SW release for next week? Thanks!

knopers8 (approved):

PR #662, opened 2024-09-18
WIP OCTRL-917 add public "bookkept" flag to enable bookkeeping of an env

knopers8:


workflows/readout-dataflow.yaml
```diff
@@ -1,5 +1,6 @@
 name: !public readout-dataflow
 description: !public "Main workflow template for ALICE data taking"
+bookkept: !public "true"
```
vascobarroso, line 3: @knopers8 does this new var need to be marked as public ? IIRC, that's to control if the var is displayed in the GUI or not as a widget.

knopers8: i assumed it could be, since I consider it a high-level property of a workflow template and it could be displayed in one way or another whether the environments started with it are bookkept.

teo: I'm not sure it should be a widget at all, and in any case, this is not the syntax that would make it a widget. It's not even clear that it's a static property of the workflow. My understanding was that it's not up to the user to choose whether an environment will be `bookkept` or not, but that it depends on some combination of traits of the workflow itself, and of its runtime configuration. In any case, I don't think this is the syntax we want.

@vascobarroso under which conditions should an environment be bookkept / not be bookkept?

teo: Are all instances of `readout-dataflow` to be bookkept, no matter the run type, config etc?

vascobarroso: I think so, that's what we do now via gRPC correct ? Or do you see a case where we don't do the gRPC call ?

teo: @vascobarroso whether the gRPC calls are triggered currently depends on `bookkeeping_enabled` which is fully runtime.

teo:
Closing in favour of https://github.com/AliceO2Group/ControlWorkflows/pull/664

PR #665, opened 2024-10-02
O2-5372 Allow ITS to accept CRU packets during PFR/EOR

knopers8:


vascobarroso:
@teo this has been tested by @pippohub on STAGING and seems to work as expected. 
One thing he mentioned is that when the basic task fails (e.g. he tested without the script being present on the FLP) the env does not go to error even if the task is critical. Maybe because the env is already destroying anyway ?  If you want to check the logs, here's a sample env id: `2qFkZfxHTBA`

teo:
> @teo this has been tested by @pippohub on STAGING and seems to work as expected. One thing he mentioned is that when the basic task fails (e.g. he tested without the script being present on the FLP) the env does not go to error even if the task is critical. Maybe because the env is already destroying anyway ? If you want to check the logs, here's a sample env id: `2qFkZfxHTBA`

Yes, I confirm this behaviour. At `DESTROY` we only log any hook failure to IL. By the time this hook runs, there is no environment state to speak of because all the tasks have been torn down already, so we don't use the critical flag at all.

teo (approved):

PR #666, opened 2024-10-13
CTP: added HI FLP workflows with QC

aferrero2707:
This adds versions of the FLP workflows for the heavy ion configuration that include the QC command.

It also adds the run and beam types to the Apricot parameters, to allow more flexibility in the QC configuration.


aferrero2707:
@vascobarroso @Barthelemy @lietava those changes are needed to enable the CTP QC for the PbPb run, and to add more flexibility with the Apricot parameters.

Barthelemy (approved):

PR #667, opened 2024-10-13
ZDC: combine local and remote QC configuration into a single file

aferrero2707:
With the current ZDC workflows and QC configuration, the ZDC FLP is not only running the raw QC task, but also the corresponding checker and post-processing task.
While this is not forbidden, it would be preferable to move the checker and post-processing to the remote QC node, and only leave the task on the FLP.

Moreover, given that ZDC has only one FLP node, it seems more practical to combine all tasks in a single multi-node configuration file, to avoid duplicating parts of the configuration in multiple files.

Two new workflows, one local and one remote, are cloned from the original ones:
- `zdc-full-qcmn-local` from `zdc-raw-qc`
- `zdc-full-qcmn-remote` from `zdc-rec-epn`

Compared to the original workflows, the new ones have the following changes:
- the parameters `--local --host flp` are added to the `o2-qc` command in the FLP workflow, to prevent the execution of the checker and post-processing tasks
- the remote QC node workflow is modified to read the combined QC configuration file

In addition, I have appended the `run_type` and `beam_type` parameters to the apricot URLs in the new workflows, to allow extra flexibility in the QC configuration if needed.

More details are given in https://its.cern.ch/jira/browse/OCTRL-941


teo (approved):

Barthelemy (approved):
This is great, thank you !

vascobarroso:
@Barthelemy @aferrero2707 this has now been merged but it's not yet in production.
Please let me know when you would like to move this into production.

PR #669, opened 2024-12-11
Sort detector list where appropriate

teo:
Requires https://github.com/AliceO2Group/Control/pull/643 and AliECS v1.26


knopers8 (approved):
i assume marshaling preserves the order...

PR #670, opened 2024-12-13
Ensure trg.RunStop happens strictly before the O² stop timestamp

teo:


vascobarroso (approved):

knopers8:
Hi @teo, what was the rationale and context behind this change?

knopers8:
@teo ping :arrow_up:

knopers8:
@vascobarroso do you know? :arrow_up:

vascobarroso:
@knopers8 during the migration of AliECS/bookkeeping communication from gRPC to kafka we noticed that the TRG stop was not happening before the O2 stop. I believe this change ensures that TRG stop is strictly before O2 stop (as it should be).

knopers8:
@vascobarroso isn't it enough if TRG stop happens before run end completion time (i.e. end of end of run)? O2 stop is the start of end of run, so to me, anything that contributes to the end of run, should happen after that.

vascobarroso:
@knopers8 what we want to ensure is that TRG is stopped before we begin the stop of O2 processes (e.g. readout) to ensure that the dataflow is terminated in the correct order. Or in other words, when we ask readout to transition from RUNNING to CONFIGURED, the trigger should already be stopped.

knopers8:
@vascobarroso Then this change was not needed. As you can see in the [documentation](https://github.com/AliceO2Group/Control/blob/master/docs/handbook/operation_order.md#stop_activity-end-of-run), negative and positive hooks for `before_STOP_ACTIVITY` are invoked before tasks are transitioned to `CONFIGURED`. For `GO_ERROR` (while not documented), it's the same.

If you saw this is not respected, it's a bug and it should be investigated.

vascobarroso:
@knopers8 indeed I did not actually see and compare the state transitions. What I saw was that the O2 stop time was sometimes earlier than the TRG stop time in bookkeeping.

PR #679, opened 2025-04-17
[dd] change RDH version from 6 to 7

vascobarroso:


vascobarroso:
@lkrcal we use RDH v7, not v6. Could you check what would be the impact of this change in StfBuilder ?

vascobarroso:
@lkrcal new ping

PR #681, opened 2025-05-16
ctp for pp high rate with 6 ,12 and 16 decoders

lietava:


lietava:
@knopers8 : thanks. It is updated.

knopers8 (approved):

PR #705, opened 2026-05-14
TPC: fix proxy channel name in tpc-cmv

ehellbar:


ehellbar:
@rmunzer

ehellbar:
I only changed the output-proxy channel name from tpcidc to tpccmv

knopers8 (approved):

PR #711, opened 2026-06-29
Default installations should not call Bookkeeping plugin

knopers8:
Bookkeeping already migrated to use ECS core kafka events. The plugin was already disabled on staging/prod setups in consul, but we missed standard one-node installations.


knopers8:
@graduta that should be sufficient as far as I know.

graduta (approved):

